PR #4, opened 2020-09-16
User/htruong/uprev piwebcam fps

### LINKED ISSUE #3: Poor performance on Cisco Webex (macOS Catalina)
Even thought the perforce of the camera is excellent in FaceTime or capturing videos from QuickTime, when I try to use it on Webex it falls drastically. Lots of dropped frames (only id say it only captures at 2-4 fps). Also, the quality of the video on the other end of the call is terrible.

I also noticed that Skype won't let me set this device as the default camera. When I select it from the menu the preferences window crashes.

Any idea how I can troubleshoot this?

htruong:
Uprev piwebcam so we can get 30fps on cisco webex. Fixes #3.

### DIFF
--- a/package/piwebcam/piwebcam.mk
+++ b/package/piwebcam/piwebcam.mk
@@ -4,7 +4,7 @@
 #
 ################################################################################
 
-PIWEBCAM_VERSION = f120aff082a03526c1327bd95f4136f08943758d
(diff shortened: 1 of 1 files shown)


htruong:
Awaiting test reports. Will merge once I receive confirmation that this release works.

PR #10, opened 2020-10-07
Add expect script for running commands over serial.

volfyd:
Hi...it would be nice to have a small expect script to interact with the CLI on the serial port.


htruong:
Thanks for the contribution!

Can we have the script take the device as a parameter?

htruong:
Thanks for the improvements! Could I request another small change, since this is a small change, can you squash all the commits together and make it one commit?

volfyd:
No problem. Done.

htruong:
Thanks! I have tested the script, it worked as expected :+1:).

PR #12, opened 2020-10-08
Support Raspberry Pi Zero v1.3 (no Wi-Fi)

tuyenld:
- We need overlays partion because of setting "dtoverlay=dwc2" in the config.txt
dwc2.dtbo will be looked for in the /boot/overlays
(see more: part "3.1: DTBs, overlays and config.txt" https://www.raspberrypi.org/documentation/configuration/device-tree.md)

- According to https://github.com/raspberrypi/firmware/blob/d46e089301ba5b2e6768c78d706e9d28370e3c1c/boot/overlays/README#L722 it read
Name:   dwc-otg
Info:   Selects the dwc_otg USB controller driver which has fiq support. This
        is the default on all except the Pi Zero which defaults to dwc2.
Load:   dtoverlay=dwc-otg
Params: <None>

This is not true. If the dwc2 wasn't use, you'll have an error:
$ journalctl --no-pager -u piwebcam
Feb 07 15:51:04 showmepicam piwebcam[153]: UVC: device open failed: No such file or directory (2).
Feb 07 15:51:04 showmepicam systemd[1]: piwebcam.service: Main process exited, code=exited, status=1/FAILURE
Feb 07 15:51:04 showmepicam systemd[1]: piwebcam.service: Failed with result 'exit-code'.

PR #13, opened 2020-10-10
pi4b support with hq camera module

osresearch:
preliminary support for building pi4b images; it works for me, although I don't have a pi0 to check that it didn't disturb anything.


build-showmewebcam.sh
```diff
@@ -1,26 +1,26 @@
...
-			sed "1i ### DO NOT EDIT, this is generated file from $ORIGIN_PIZEROW_CAM defconfig" -i "$target_defconfig_loc"
+			sed "1i ### DO NOT EDIT, this is generated file from $BOARD_NAME defconfig" -i "$target_defconfig_loc"
 			# change path
 			sed "s/board\/raspberrypi0cam/board\/$1/g" -i "$target_defconfig_loc"
 			# change dts file
 			sed "s/bcm2708-rpi-zero-w/bcm2708-rpi-zero/g" -i "$target_defconfig_loc"
             ;;
         *)
-			target_defconfig="$ORIGIN_PIZEROW_CAM"_defconfig
+			target_defconfig="$BOARD_NAME"_defconfig
```
tuyenld, line 21:
```suggestion
			target_defconfig="$BOARD_NAME"_defconfig
```
In the current version, `./build-showmewebcam.sh` is used to build image for Pi Zero W. I think we should keep Pi Zero W as default because of backward compatibility.

configs/raspberrypi4cam_defconfig
```diff
@@ -0,0 +1,47 @@
...
+BR2_TARGET_GENERIC_GETTY_PORT="ttyAMA0"
+BR2_ROOTFS_OVERLAY="$(BR2_EXTERNAL_PICAM_PATH)/board/raspberrypi0cam/rootfs"
+BR2_ROOTFS_POST_BUILD_SCRIPT="$(BR2_EXTERNAL_PICAM_PATH)/board/raspberrypi0cam/post-build.sh"
+BR2_ROOTFS_POST_IMAGE_SCRIPT="$(BR2_EXTERNAL_PICAM_PATH)/board/raspberrypi0cam/post-image.sh"
+BR2_ROOTFS_POST_SCRIPT_ARGS="--configure-picam"
+BR2_LINUX_KERNEL=y
+BR2_LINUX_KERNEL_CUSTOM_TARBALL=y
+BR2_LINUX_KERNEL_CUSTOM_TARBALL_LOCATION="$(call github,raspberrypi,linux,4f2a4cc501c428c940549f39d5562e60404ac4f7)/linux-4f2a4cc501c428c940549f39d5562e60404ac4f7.tar.gz"
+BR2_LINUX_KERNEL_USE_CUSTOM_CONFIG=y
+BR2_LINUX_KERNEL_CUSTOM_CONFIG_FILE="$(BR2_EXTERNAL_PICAM_PATH)/board/raspberrypi0cam/linux-$(BOARD_NAME).config"
```
tuyenld, line 19:
```suggestion
BR2_ROOTFS_POST_BUILD_SCRIPT="$(BR2_EXTERNAL_PICAM_PATH)/board/raspberrypi0cam/post-build.sh"
BR2_ROOTFS_POST_IMAGE_SCRIPT="$(BR2_EXTERNAL_PICAM_PATH)/board/raspberrypi0cam/post-image.sh"
BR2_LINUX_KERNEL_CUSTOM_CONFIG_FILE="$(BR2_EXTERNAL_PICAM_PATH)/board/raspberrypi0cam/linux-$(BOARD_NAME).config"
```
`post-image.sh` use basename to build an appropriate image. For example
raspberrypi0cam/post-image.sh -> build image for **raspberrypi0cam** which define in raspberrypi0cam_defconfig
raspberrypi0/post-image.sh        -> build image for **raspberrypi0**        which define in raspberrypi0_defconfig
I think it is more understandable.

I see raspberrypi4cam_defconfig is slightly different from raspberrypi0cam_defconfig, what do you think if adding these changes programmatically  in build.sh?

osresearch: The `raspberrypi4cam_defconfig` is derived from the `bcm2711_defconfig`, which is the default kernel configuration for the Pi4. I merged some of the changes from the pi0cam config; perhaps it would be better to have the changes relative to the board specific configs?

build-showmewebcam.sh
osresearch: Yes, I'll change the default back.

configs/raspberrypi4cam_defconfig
tuyenld: @osresearch
That is what I did on this [commit](https://github.com/showmewebcam/showmewebcam/pull/12/commits/3bcbe30fc8d8a3badf3629f9c8272bb80b61587c#).
I copy raspberrypi0cam_defconfig, then change path and dts file.
How do you think about this approach?

PR #22, opened 2020-10-26
Mark Pi Zero v1.3 (without Wi-Fi) as tested for HQ camera

Leonti:
My HQ camera finally arrived today and I was able to test the setup with RPi Zero (non-WiFi) version.
I used `v1.40` release (`sdcard-raspberry0-v1.40.img.gz`) and the camera works as expected.
Many thanks!


htruong:
Thank you for verifying and submitting the patch!

PR #24, opened 2020-11-03
disable root fsck on boot

dgsiegel:
Removes some useless warnings as we use squashfs anyway


htruong:
Thanks for sending this in @dgsiegel!

One thing I noticed is that the boot time has not been improved. Not that it matters, but it would expect it to be improved. Any idea why?

htruong:
Before:

```
# systemd-analyze
Startup finished in 2.109s (kernel) + 9.647s (userspace) = 11.756s
multi-user.target reached after 9.291s in userspace
```

After:

```
# systemd-analyze
Startup finished in 2.113s (kernel) + 9.609s (userspace) = 11.723s
multi-user.target reached after 9.229s in userspace
```

dgsiegel:
It shouldn't really improve boot performance, but remove the fsck warning on boot

PR #25, opened 2020-11-04
disable unused systemd services

dgsiegel:
This should shave off a few milliseconds...

### DIFF
--- a/board/raspberrypi0cam/post-build.sh
+++ b/board/raspberrypi0cam/post-build.sh
@@ -23,3 +23,7 @@ fi
 
 # Disable fsck on root
 sed -ie '/^\/dev\/root/ s/0 1/0 0/' "${TARGET_DIR}/etc/fstab"
+
+# Disable unused services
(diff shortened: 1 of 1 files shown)


board/raspberrypi0cam/post-build.sh
```diff
@@ -23,3 +23,7 @@ fi
 
 # Disable fsck on root
 sed -ie '/^\/dev\/root/ s/0 1/0 0/' "${TARGET_DIR}/etc/fstab"
+
+# Disable unused services
+ln -sf /dev/null "${TARGET_DIR}"/etc/systemd/system/sys-kernel-debug.mount
```
htruong, line 28: I'm a little bit curious about your choice of symlinking the file to `/dev/null`. I'm not well versed in the systemd land, but why didn't you just remove it?

peterbay: Symlink to /dev/null is used for masking - This is the third level of “off” in systemd. If you boot with a unit masked, it will not run even to satisfy dependencies. 
[systemd: Masking units](https://fedoramagazine.org/systemd-masking-units/)

PR #26, opened 2020-11-04
use latest raspberry pi kernel

dgsiegel:
We'll need this to get the latest version of uvc, which fixes a few issues related to setting multiple resolutions


htruong:
Verified that the kernel is uprevved. Thanks!

```
# uname -a
Linux piwebcam 5.4.73 #1 Mon Nov 9 22:26:29 PST 2020 armv6l GNU/Linux
```

PR #29, opened 2020-11-04
optimize boot time

dgsiegel:
This includes:
- removing dbg/printk/trace functions in the kernel
- adding lz4 compression to the kernel
- setting the bootloader delay to 0 seconds
- informing the kernel about the filesystem and ro state on boot time
- suppress kernel output during boot

Before:

Startup finished in 3.351s (kernel) + 8.779s (userspace) = 12.131s 
multi-user.target reached after 8.469s in userspace

After:

Startup finished in 2.060s (kernel) + 6.776s (userspace) = 8.837s 
multi-user.target reached after 6.459s in userspace

About 4-5 seconds :)


dgsiegel:
Current boot sequence:

```
2.726s dev-mmcblk0p2.device                  
1.864s systemd-udev-trigger.service          
1.014s systemd-journald.service              
 829ms systemd-udevd.service                 
 656ms systemd-modules-load.service          
 610ms systemd-update-utmp.service           
 530ms network.service                       
 324ms systemd-journal-catalog-update.service
 302ms systemd-remount-fs.service            
 295ms tmp.mount                             
 276ms sys-kernel-config.mount               
 261ms systemd-sysctl.service                
 222ms systemd-update-utmp-runlevel.service  
 148ms boot.mount                            
 141ms var.mount                             
 134ms systemd-journal-flush.service         
  82ms systemd-update-done.service       
```

htruong:
Amazing results. Thanks for the PR.

Issue #23, opened 2020-11-03
Add camera.txt to /boot during build


dgsiegel:
It would be awesome if an existing camera.txt could be copied to /boot when building an image yourself. With that in place, we could also remove 

```
/usr/bin/v4l2-ctl -c auto_exposure_bias=3
/usr/bin/v4l2-ctl -c video_bitrate=25000000
```
from `package/piwebcam/start-webcam.sh`. The only solution I found would be to add the camera.txt file to `board/raspberrypi0cam/genimage-raspberrypi0.cfg` and add these lines to `board/raspberrypi0cam/post-image.sh`:

```
# Copy camera.txt if it exists in build dir
if [ -f "${BR2_EXTERNAL_PICAM_PATH}/camera.txt" ]; then
  cp "${BR2_EXTERNAL_PICAM_PATH}/camera.txt" "${BINARIES_DIR}/rpi-firmware/camera.txt"
fi
```

But that requires an existing camera.txt in the root folder or the build will fail.

htruong:
Not that I tested this idea, but we could put a dummy blank `camera.txt` file in the repo so that it doesn't upset the `cfg` file config, and then somehow dynamically swap it to the real file to it during build, maybe?

dgsiegel:
That would work if we add the above mentioned camera.txt file to the repo. Don't know if that is the best way though...

PSF1:
The repo has a camera.txt.example in it. That file have configuration details but it don´t change default values. it maybe copied how a empty camera.txt file while building.

dgsiegel:
Well, we already have these two lines in `package/piwebcam/start-webcam.sh`:
```

/usr/bin/v4l2-ctl -c auto_exposure_bias=3
/usr/bin/v4l2-ctl -c video_bitrate=25000000
```

They could easily go into the default camera.txt, no?

dgsiegel:
Here's a pull request for this: https://github.com/showmewebcam/showmewebcam/pull/31

rcarmo:
It would be nice to have a sample `camera.txt` by default with the most common settings there (including rotation, which is what I need to tweak on every new build...).

dgsiegel:
Yeah, we have some settings already (https://github.com/showmewebcam/showmewebcam/issues/23#issuecomment-723344994). What other settings would you propose?

Regarding rotation, that probably depends on how you use your mount?

rcarmo:
[...]

PR #31, opened 2020-11-10
add default camera.txt which can be overwritten during build

dgsiegel:
Create a default camera.txt and allow to overwrite it during build time by adding a camera.txt file to the root of the repository

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 output
 board/raspberrypi0
 configs/raspberrypi0_defconfig
+/camera.txt
--- a/README.md
+++ b/README.md
@@ -111,6 +111,7 @@ Make a directory in your home: `develop`.
   - `BUILDROOT_DIR=../buildroot ./build-showmewebcam.sh raspberrypi0w` to build Raspberry Pi Zero W (with Wi-Fi) image.
   - `BUILDROOT_DIR=../buildroot ./build-showmewebcam.sh raspberrypi0` to build Raspberry Pi Zero (without Wi-Fi) image.
 - The resulting image will be at the `output/` folder in the showmewebcam directory.
+- If you add a camera.txt file to the root of this repository, the contents will be automatically added to /boot/camera.txt
 
 Credits
 --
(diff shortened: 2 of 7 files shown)


tuyenld:
Hi, @dgsiegel,

I am thinking about the way we can create/edit `camera.txt` directly from uvc-gadet. Users don't need to create or edit it manually.
Here is an idea:
1. uvc-gadget support abundance camera control ([link](https://github.com/peterbay/uvc-gadget/blob/master/doc/src/video-controls.md)), but we need to enable these setting on the linux kernel
2. Currently, there is no configuration file setting for uvc-gadget. one for these above settings is needed.
3.  `/boot` partition is read-only, how can we write a configuration file. I think we can watch a temporary configuration file, then use [inotifywait](https://linux.die.net/man/1/inotifywait) to remount and write it to `camera.txt`.

dgsiegel:
I think we should provide a default camera.txt file which works for most cameras/lenses, see https://github.com/showmewebcam/showmewebcam/issues/23#issuecomment-725421478

As for editing, I disagree, as some lenses (once you go into the c/cs-mount lens land) need some configuration to look better. 

Then, as uvc-gadget supports camera control on the host computer as well, users are free to edit either the camera.txt or use v4l2-ctl on their computer.

htruong:
> Hi, @dgsiegel,
> 
> I am thinking about the way we can create/edit `camera.txt` directly from uvc-gadet. Users don't need to create or edit it manually.
> Here is an idea:
> 
> 1. uvc-gadget support abundance camera control ([link](https://github.com/peterbay/uvc-gadget/blob/master/doc/src/video-controls.md)), but we need to enable these setting on the linux kernel
> 2. Currently, there is no configuration file setting for uvc-gadget. one for these above settings is needed.
> 3. `/boot` partition is read-only, how can we write a configuration file. I think we can watch a temporary configuration file, then use [inotifywait](https://linux.die.net/man/1/inotifywait) to remount and write it to `camera.txt`.

We can certainly do that. Once peterbay's uvc-gadget can change the parameters on the fly, we can create a daemon or something that watches the configuration as you said, if it changes then it can remounts `/boot` and writes the new config in.

PR #32, opened 2020-11-10
update uvc-gadget

dgsiegel:
this gives us better performance as well as led display when the camera is on


htruong (commented):
Very nit: Uvc-gadget instead of uvc-widget.

dgsiegel:
> Very nit: Uvc-gadget instead of uvc-widget.

🤦‍♂️️

htruong:
Smoke tested and feature works as expected!

Issue #9, opened 2020-10-04
Field of views problems 


Fernyman79:
I'm using the raspberry pi v2 camera module. When testing the camera with raspistill, I get to see the full field of view. However, when I load the showmewebcam image on the Pi, the video is cropped. Some sample images:

> raspistill
![raspistill](https://user-images.githubusercontent.com/44389280/95007958-3bd0bb80-05ca-11eb-9f27-35ad9be29bcb.jpg)

> showmewebcam
![showmewebcam](https://user-images.githubusercontent.com/44389280/95007960-3ffcd900-05ca-11eb-9b1a-d688de9a5320.jpg)

I forgot to mention on my other thread: Thanks for making this! Trully awesome!

towolf:
Same here, it's really strongly cropped.

PSF1:
Same here. Pi 0w + 1.3 cam

towolf:
I think it's because of this table:

https://picamera.readthedocs.io/en/release-1.13/fov.html#sensor-modes

I would like 640x480, with 4x4 binning. Good quality and full view.

volfyd:
FYI, the modes for the HQ camera (and also V1 and V2) are here: https://www.raspberrypi.org/documentation/raspbian/applications/camera.md

I wonder if you could get the mode you wanted by changing the values in https://github.com/showmewebcam/showmewebcam/blob/master/package/piwebcam/multi-gadget.sh

For example, s/1080/480/ and s/1920/640/

If that works, maybe the script could look for a resolution override file in /boot

htruong:
@towolf THanks for the excellent resource. I will take a look at it and see if changing the resolution solves the issue.

volfyd:
> FYI, the modes for the HQ camera (and also V1 and V2) are here: https://www.raspberrypi.org/documentation/raspbian/applications/camera.md
> 
> I wonder if you could get the mode you wanted by changing the values in https://github.com/showmewebcam/showmewebcam/blob/master/package/piwebcam/multi-gadget.sh
> 
> For example, s/1080/480/ and s/1920/640/
> 
> If that works, maybe the script could look for a resolution override file in /boot

For anyone trying this themselves, it seems that the resolution is also hard coded in uvc-gadget.c which is downloaded by the buildroot process somehow.

htruong:
[...]

PR #33, opened 2020-11-11
Fix Field of views problems #9

tuyenld:
- Avoid duplicate with 1920x1080 which also create /sys/kernel/config/usb_gadget/piwebcam/functions/uvc.usb0/streaming/mjpeg/m/1080p
- 1440x1080 will have pixels wide stripe on the right side (the reason is unknown)


tuyenld:
I tested with Pi Zero v1.3 + Pi Cam v1.
I not sure if it'll work with HQ Camera.

htruong:
Yeah, I'd rather wait to understand the root cause of this issue rather than merge something that we don't know why it works.

PR #35, opened 2020-11-12
optimize boot time pt. 2

dgsiegel:
This includes:
- removed debug capabilities from kernel
- removed unneeded kernel modules
- removed network service if it's a raspberry 1.3
- removed utmp systemd services

We're down to:
Startup finished in 784ms (kernel) + 5.706s (userspace) = 6.491s 
multi-user.target reached after 5.677s in userspace

@htruong would you mind testing this with the Raspberry Zero W to check if wifi still works?


dgsiegel:
Together with a few final tweaks, like `initial_turbo` we're now under 5 seconds

Startup finished in 657ms (kernel) + 4.079s (userspace) = 4.736s 
multi-user.target reached after 4.063s in userspace

htruong:
Sure, I will test tomorrow and will merge if it passes the smoke test!

htruong:
Works for me:

```
Startup finished in 758ms (kernel) + 5.562s (userspace) = 6.321s
multi-user.target reached after 5.501s in userspace
```

PR #36, opened 2020-11-16
Adapt to changed hostname and hence prompt

towolf:


dgsiegel:
Merged. Thanks so much :)

PR #38, opened 2020-11-20
Control GPIO pin 21 to indicate when camera is streaming

Leonti:
Use functionality implemented in `uvc-gadget` to control pin 40 (GPIO 21) which can be used for an external LED or as an input to trigger other things (a ring light, an "ON-AIR" sign, etc.)


Leonti:
This commit has been built and tested on my Raspberry Pi Zero which has a green 5mm LED

dgsiegel:
Hey, thanks for the pull request. What implications does this have, if you either don't have any GPIO pins, or leave pin 40 empty?

Leonti:
Every Raspberry Pi has 40 GPIO pins, so the pin will definitely be there.
If nothing is connected to the pin it has no effect - it just changes the state of the pin.

CONFIG_GPIO_SYSFS is needed to enable controlling GPIO through sysfs which is used by `uvc-gadget`:
https://github.com/peterbay/uvc-gadget/blob/master/uvc-gadget.c#L76

Leonti:
By the way, pin 21 was chosen because it's the outermost pin residing in a corner, so it's easy to find even without having a pinout handy - very convenient for soldering :)

dgsiegel:
Ok, let me try with my GPIO-less Pi. In the mean time, would you mind adding that information to the README as well? Might make sense to add a section about the internal LED as well.

dgsiegel:
Just tested it on 1.3 without GPIO pins and looks great. If you could just add one or two lines to the README about this, we can go ahead and merge this.

Leonti:
Thanks for looking into it!
I've updated README with a short explanation about built-in LED and GPIO 21.

dgsiegel:
Awesome, thanks!

PR #39, opened 2020-11-21
add github action to build an image after each commit

dgsiegel:
Unfortunately Github doesn't allow public access to build artifacts, but it's the best we can do right now if we don't want to mess with dev-releases.

- The initial build takes about 1.5 hours. The output directory will be cached and subsequent builds should be less than 5 minutes.
- Kernel was switched back to tarball, to reduce download and storage size
- The -s flag for make reduces build verbosity and makes the build logs easier to read


htruong:
This is the hero we need!

PR #40, opened 2020-11-22
Automatically compress image after building

htruong:
To save github some space and bandwidth and to help with our
unwillingness to have to compress the images manually, automatically
compress the resulting image after building.

Also, change the copy directive from the workflow to a move, it's faster
to move than to copy.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -49,10 +49,10 @@ jobs:
       - name: Build image
         run: |
           sh build-showmewebcam.sh ${{ matrix.raspberry }}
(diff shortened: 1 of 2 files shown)


dgsiegel:
All assets will be automatically compressed by Github. You can check it yourself by downloading an artifact. With your proposal, you'll find the images as e.g. "sdcard-raspberrypi0w-dd9c949.img.gz.zip", hence double-zipped :)

The mv command would be fine however

htruong:
Stupid question, can the programs like etcher deal with zipped (not gzipped) images?

dgsiegel:
Not that I know of. Most programs on Linux use dd in some way or another, which expects plain images.

htruong:
Ah, I just tried. It actually can flash image-inside-a-zip files fine. So we actually don't need this. I will change the cp to mv at some other point when it's convenient.

PR #41, opened 2020-11-22
Tweak default camera.txt settings

htruong:
According to testing results in #23

### DIFF
--- a/package/piwebcam/camera.txt
+++ b/package/piwebcam/camera.txt
@@ -3,5 +3,8 @@
 # Use v4l2-ctl -L or have a look at camera.txt.example
 # to view all possible settings
 
-auto_exposure_bias=3
+auto_exposure_bias=12
+contrast=-2
+saturation=5
 video_bitrate=25000000
+sharpness=20


package/piwebcam/camera.txt
```diff
@@ -3,5 +3,9 @@
 # Use v4l2-ctl -L or have a look at camera.txt.example
 # to view all possible settings
 
-auto_exposure_bias=3
+auto_exposure=0
```
dgsiegel, line 6: ` auto_exposure=0` seems to be the default. Should we drop it?

package/piwebcam/camera.txt
```diff
@@ -3,5 +3,9 @@
...
+auto_exposure_bias=12
```
dgsiegel, line 7: I think we could use 13 or 14 here as well? What do you think?

htruong: This is "good enough" for me.

package/piwebcam/camera.txt
htruong: Changed it with newest push!

PR #44, opened 2020-11-25
Build universal Raspberry Pi 0 image

htruong:
Instead of having to build two separate images, we can have one universal
image with the two dtb files for both Raspberry Pi 0 variants.


dgsiegel:
It seems to work on my Zero 1.3. But I am wondering if this is a good idea :) It would only make our Github Actions script and our release easier. But when we think about the Pi 3 or Pi 4, we could certainly add support for them as well (See #13).

Additionally I've removed some network functions on the 1.3, which would increase the boot time by .5s at least.

Do you have any arguments for it?

htruong:
I think having one universal image is advantageous in at least three ways:

- Less testing for us (if the universal image works, we will assume it will work on both platforms)
- Less build time for us.
- Less confusion for the end-users.

I will fix the wireless part... I think the best course of action is to remove networking altogether, so the cam can't possibly communicate outside bypassing the host computer in any way. That increases trust in our fw.

dgsiegel:
It's a pity if one has a 0w, that they can't use their wifi anymore. But it's understandable from a privacy point of view.

If we add support for the Pi4, we would need different architectures anyway. Build time is negligent as they will be built in parallel here...

htruong:
If we really want network, then we can always detect what pi variant we're running on at runtime, no need for build-time distinction.

dgsiegel:
Is there some information available what the difference between the individual firmware files is? Apparently `bcm2708-rpi-b-plus.dtb` includes support for all Zero variants...

> If we really want network, then we can always detect what pi variant we're running on at runtime, no need for build-time distinction.

In that case we should remove networking support from the kernel as well?

htruong:
> Apparently `bcm2708-rpi-b-plus.dtb` includes support for all Zero variants...

Do you have a source for that?



> In that case we should remove networking support from the kernel as well?

I'm not opposed to having network-over-USB, because it requires host computer to initiate the connection and doesn't expose us to anything more than the tty-over-usb already exposes.

dgsiegel:
> Do you have a source for that?

Nope, just an entry in a forum unfortunately: https://www.raspberrypi.org/forums/viewtopic.php?p=1456353&sid=984070cd1a59069c13a0250ff2e1a973#p1456353

htruong:
>A little known fact: if running on a Pi Zero, the firmware will look for bcm2708-rpi-zero.dtb, falling back to bcm2708-rpi-b-plus.dtb if it doesn't find it, so you could rename bcm2708-rpi-b-plus.dtb for your Zero-dedicated cut-down image.

I have tried to look for the source for the claim, but I think bootcode.bin/stat.elf is still closed-source -- so no way we can verify that. I think there is no sense to not use the preferred dtb file.

htruong:
I would wait until the Pi4 support merges first, no need to delete a bunch of support code and make it hard for the Pi4 support now.

htruong:
Declining this in favor of https://github.com/showmewebcam/showmewebcam/pull/173

PR #48, opened 2020-11-26
Build releases automatically when pushing tags

dgsiegel:
This will add a new tag and build a release on github if you push any annotated tag to the upstream repository. Additionally all commits since the last tag will be added to the release notes for further editing.

Lastly, we move the ccache directory to inside output/ in order to cache it as well. The downside obviously is that you'll loose the ccache history when you clean your build directory.


.github/workflows/release.yml
```diff
@@ -21,42 +23,106 @@ jobs:
...
           key: ${{ matrix.raspberry }}-${{ env.BUILDROOT_DIR }}-${{ env.BUILDROOT_VERSION }}-${{ env.CACHE_VERSION }}
 
-      - name: Build image
+      - name: Build Image
         run: |
+          # Force a piwebcam rebuild every time
+          rm -fr output/*/build/piwebcam*
           sh build-showmewebcam.sh ${{ matrix.raspberry }}
-          mv output/${{ matrix.raspberry }}/images/sdcard.img sdcard-${{ matrix.raspberry }}-${{ env.SHA }}.img
+          cp output/${{ matrix.raspberry }}/images/sdcard.img sdcard-${{ matrix.raspberry }}-${{ env.SHA }}.img
```
htruong, line 65: are we going back to `cp` now, why?

configs/raspberrypi0cam_defconfig
```diff
@@ -1,6 +1,7 @@
 BR2_arm=y
 BR2_arm1176jzf_s=y
 BR2_CCACHE=y
+BR2_CCACHE_DIR="$(BASE_DIR)/.ccache"
```
htruong, line 4: Nice!

.github/workflows/release.yml
dgsiegel: yeah, because we need that file for the release. i thought it's easier to always use the base image for the right naming.

PR #49, opened 2020-11-27
Raspberry Pi4 Support

skarard:
I gave #13 (pi4b support with hq camera module) a go but it didn't boot for me. 

It felt worth breaking out board/raspberrypi4cam as a separate folder. The rationale behind this is that the additional power of Pi4 may be able to support more advanced image and tensor flow operations. Rather than sedding the diff I wanted the flexibility to configure in place, it's more readable and allows for comments.

build-showmewebcam.sh got a bit crazy because I was unclear on the direction of the generated board configuration path. 

Awesome work putting this all together!


skarard:
`ERROR: file(rpi-firmware/fixup4.dat): stat(/home/runner/work/showmewebcam/showmewebcam/output/raspberrypi4cam/images/rpi-firmware/fixup4.dat) failed: No such file or directory`

I'm not entirely sure why this errored out. I built this locally using the same version of buildroot (2020.02.3) and the correct files were generated with this name in this location. I'll see if a recent merge for the build environment changed anything.

skarard:
For anyone wanting a build for the Pi 4, below is a temporary release until this merged in properly.

https://github.com/skarard/showmewebcam/releases/tag/v1.60-pi4-alpha

dgsiegel:
Thanks so much for doing this! I'd definitely like to merge this for the next release.

Here's a few pointers:
- Currently we don't store the other variants as config files, but link and copy the existing configuration over. This is done on the fly in `build-showmewebcam.sh`. If you have any better idea feel free to propose something, otherwise we'd like to stick to the current behaviour
- We are currently in the process of moving everything over to buildroot-2020.08.2 (already used in our Github actions and releases), please make us of that as well
- Please use `board/raspberrypi0cam/genimage-raspberrypi4.cfg` as the main path for your config
- Can't really track the changes in the `linux.config`. Are there any changes in there required for the Pi4?
- Same for `post-image.sh`/`post-build.sh `

htruong:
Agreed with @dgsiegel 

Having the pi4 support now is very nice, and I would like to ask you to do a step further and have it in a way that makes it easy to maintain. However, I would much prefer that the kernel config for the pi4 is in the diff format on top of what's selected for the pi0. That way we can be sure any enhancements to pi0 are applicable to the pi4 and vice-versa. In the current format, we don't know what is selected for the pi4 that the pi0 doesn't have selected.

We would need something like a config wrapper for the pi4 so that it applies the diff for the pi4 kernel|menu config on top of each time we do a pi4 config.

skarard:
How did you generate board/raspberrypi0cam/linux.config? 

I have attempted to manually change yours and update it with the minimal requirements for Pi4 but the build won't boot. 
So I switched tack and used the default Pi4 .config from buildroot. A diff comes out to over 1000 lines so it's not clear which packages are required over optional for this requirement.

I then opened your linux.config with makeconfig and i386 architecture was selected which makes no sense to me.

I concluded I needed more info as to your process for selecting your linux.config.

dgsiegel:
> How did you generate board/raspberrypi0cam/linux.config?

Our `linux.config` gets merged with the default buildroot/raspberry kernel settings. Hence we only added the needed settings there, starting with the buildroot example as a base. In any case, we'd like to reduce duplicate settings as much as possible.

> I have attempted to manually change yours and update it with the minimal requirements for Pi4 but the build won't boot.
> So I switched tack and used the default Pi4 .config from buildroot. A diff comes out to over 1000 lines so it's not clear which packages are required over optional for this requirement.

You could go the other way round and check which of our settings are different in the default Pi4 config?

htruong:
> You could go the other way round and check which of our settings are different in the default Pi4 config?

One way one could go about doing that is to inspect the kernel defconfig files in buildroot and see the secret sauces:

```
$ sdiff configs/raspberrypi0w_defconfig configs/raspberrypi4_defconfig 
BR2_arm=y							BR2_arm=y
BR2_arm1176jzf_s=y					     |	BR2_cortex_a72=y
BR2_ARM_EABIHF=y					     |	BR2_ARM_FPU_NEON_VFPV4=y
							     >
							     >	BR2_TOOLCHAIN_BUILDROOT_CXX=y
							     >
							     >	BR2_SYSTEM_DHCP="eth0"

# Linux headers same as kernel, a 4.19 series			# Linux headers same as kernel, a 4.19 series
BR2_PACKAGE_HOST_LINUX_HEADERS_CUSTOM_4_19=y			BR2_PACKAGE_HOST_LINUX_HEADERS_CUSTOM_4_19=y

BR2_TOOLCHAIN_BUILDROOT_CXX=y				     <
							     <
BR2_LINUX_KERNEL=y						BR2_LINUX_KERNEL=y
BR2_LINUX_KERNEL_CUSTOM_TARBALL=y				BR2_LINUX_KERNEL_CUSTOM_TARBALL=y
BR2_LINUX_KERNEL_CUSTOM_TARBALL_LOCATION="$(call github,raspb	BR2_LINUX_KERNEL_CUSTOM_TARBALL_LOCATION="$(call github,raspb
BR2_LINUX_KERNEL_DEFCONFIG="bcmrpi"			     |	BR2_LINUX_KERNEL_DEFCONFIG="bcm2711"
							     >
							     >	# Build the DTB from the kernel sources
BR2_LINUX_KERNEL_DTS_SUPPORT=y					BR2_LINUX_KERNEL_DTS_SUPPORT=y
BR2_LINUX_KERNEL_INTREE_DTS_NAME="bcm2708-rpi-zero-w"	     |	BR2_LINUX_KERNEL_INTREE_DTS_NAME="bcm2711-rpi-4-b"
							     >
BR2_LINUX_KERNEL_NEEDS_HOST_OPENSSL=y				BR2_LINUX_KERNEL_NEEDS_HOST_OPENSSL=y

BR2_PACKAGE_RPI_FIRMWARE=y					BR2_PACKAGE_RPI_FIRMWARE=y
							     >	BR2_PACKAGE_RPI_FIRMWARE_VARIANT_PI4=y

# Required tools to create the SD image				# Required tools to create the SD image
BR2_PACKAGE_HOST_DOSFSTOOLS=y					BR2_PACKAGE_HOST_DOSFSTOOLS=y
BR2_PACKAGE_HOST_GENIMAGE=y					BR2_PACKAGE_HOST_GENIMAGE=y
BR2_PACKAGE_HOST_MTOOLS=y					BR2_PACKAGE_HOST_MTOOLS=y

# Filesystem / image						# Filesystem / image
BR2_TARGET_ROOTFS_EXT2=y					BR2_TARGET_ROOTFS_EXT2=y
BR2_TARGET_ROOTFS_EXT2_4=y					BR2_TARGET_ROOTFS_EXT2_4=y
BR2_TARGET_ROOTFS_EXT2_SIZE="120M"				BR2_TARGET_ROOTFS_EXT2_SIZE="120M"
# BR2_TARGET_ROOTFS_TAR is not set				# BR2_TARGET_ROOTFS_TAR is not set
BR2_ROOTFS_POST_BUILD_SCRIPT="board/raspberrypi0w/post-build. |	BR2_ROOTFS_POST_BUILD_SCRIPT="board/raspberrypi4/post-build.s
BR2_ROOTFS_POST_IMAGE_SCRIPT="board/raspberrypi0w/post-image. |	BR2_ROOTFS_POST_IMAGE_SCRIPT="board/raspberrypi4/post-image.s
BR2_ROOTFS_POST_SCRIPT_ARGS="--add-miniuart-bt-overlay"		BR2_ROOTFS_POST_SCRIPT_ARGS="--add-miniuart-bt-overlay"
```

skarard:
Is it worth formatting raspberrypi0cam_defconfig to match the default raspberrypi0w_defconfig with comments for non-default options?

htruong:
This PR LGTM

skarard:
Great, I'm on the right track. 

I must have loaded the wrong image because I tried this build again and it didn't boot. I'll enable TTY on the HDMI and see what's going on. I'll let you know when it's tested.

dgsiegel:
This looks good to me as well, let us know about your progress.

Another idea: If there will be too many `sed` calls and it gets too difficult to follow your changes, did you check if it would be possible to append different config files, where the latter overwrites the former config settings?

e.g. in the default file we have `BR2_arm1176jzf_s=y` and in the pi4 config file we have `BR2_cortex_a72=y`, what happens if you merge these two files? (Maybe even inserting `BR2_arm1176jzf_s=n` as well). That way we could have a default config (both for buildroot as well as kernel) and a shorter variant for different architectures which just gets appended.

Oh and one small thing: Could you use the latest Rpi kernel and headers as we do in the default config?

skarard:
> Oh and one small thing: Could you use the latest Rpi kernel and headers as we do in the default config?

Unfortunately, the 5.4 kernel is broken for Pi 4 OTG and will not work in it's current state.

skarard:
> Another idea: If there will be too many `sed` calls and it gets too difficult to follow your changes, did you check if it would be possible to append different config files, where the latter overwrites the former config settings?

I don't know buildroot, or kernel building very well at all. I'm not sure how it reacts if we append conflicting settings. If it handles it by the last read setting overwrites a previous setting (which is how I assume it works), there are some options. My preference is to copy the default defconfig for the target device from buildroot and append the showmewebcam settings. 
Although for now dealing with the defconfig is pretty easy to sed in the substitutes.

The much more difficult case is the linux.config file. The proposed method above would work as well. Taking the default .config for that target and append showmewebcams changes.

The reason I originally supplied an entire linux.config for the pi4 was that there were 900 lines of changes. I diffed the diff out of it and thought I could get away with minimal changes, that was not the case. Now I think I've got it down to 300 lines of changes but that's not sensible to sed. I don't think it's sensible to merge, as mentioned about appending specific showmewebcam settings would be the most readable.

At the moment the linux.config file is 1241 lines are these all changes to the default to build raspiberrypi0w?

jeaton2411:
Using the Alpha.  Booted and ran well on Pi 4b when connected to PC.  On Mac, scrambled horizontal bars unless I loaded PhotoBooth which gave clear image in app and allowed other apps to use the feed but with very significant lagging.  Same result on 2014 Mac Mini AND 2014 MacBook Air.

dgsiegel:
That's good news! Unfortunately I can't help you test, but it would be great if you could make it work on Mac as well.

I was a bit confused about the many changes needed to add new targets. Hence I've pushed a change today, which hopefully makes it easier to add new targets. Here's what's changed and what we have to do to add Pi4:

- `configs/config` is our new base config for all things related to Buildroot
- You can add new targets by adding or changing the default configuration with a new file called `configs/$target`, e.g. `configs/raspberrypi0`
- Both will be merged by `build-showmewebcam.sh` into `configs/$target_defconfig`, e.g. `configs/raspberrypi0_defconfig`
- The same happens to the kernel config, only that the names will be `board/linux-base.config` and `board/linux-$target.config`. The merged result will be `board/linux.config`
- Lastly you need a `board/genimage-$target.cfg` which you already have

Now, please make sure that the config fragments needed for the Pi4 support will be at a minimum. Additionally, you can certainly add settings to the both main configs, provided they make sense as a default config. Otherwise add them to the target config.

Sorry for the additional trouble, but I think this will lead to a better result and something we can support more easily in the future.

skarard:
Thanks @dgsiegel. This is a big help. 

I am building the kernel for using this new method right now.

Am I right in thinking `linux-base.config` is based on the buildroots linux configuration for the pi0w? The modifications you made were to add Z4L kernel compression and remove packages to improve the boot time?

This means if I take the diff of buildroots pi4 and pi0w `.configs` I will be left with only the changes required to support pi4?

What happened to `BR2_arm=y` and `BR2_arm1176jzf_s=y` from `configs/config`. It looks like it's building x86.

dgsiegel:
> Am I right in thinking `linux-base.config` is based on the buildroots linux configuration for the pi0w? The modifications you made were to add Z4L kernel compression and remove packages to improve the boot time?

Correct. 

> This means if I take the diff of buildroots pi4 and pi0w `.configs` I will be left with only the changes required to support pi4?

Theoretically yes. But there might be dependencies which the diff will include but aren't needed. In that regard we could also remove or change settings from `linux-base.config` if needed.

> What happened to `BR2_arm=y` and `BR2_arm1176jzf_s=y` from `configs/config`? It looks like it's building x86.

It's right there isn't it? https://github.com/showmewebcam/showmewebcam/blob/058f6d968cc3258f1479f0edadb9570d606e9685/configs/config#L1

skarard:
> Theoretically yes. But there might be dependencies which the diff will include but aren't needed. In that regard we could also remove or change settings from `linux-base.config` if needed.

Is this something you are happy to include as a separate merge request. In a kernel minimization ticket?

> It's right there isn't it?
My bad, something weird happened with my merge.

dgsiegel:
> Is this something you are happy to include as a separate merge request. In a kernel minimization ticket?

Sure, provided we can merge it at the same time as this one.

kbingham:
> > Oh and one small thing: Could you use the latest Rpi kernel and headers as we do in the default config?
> 
> Unfortunately, the 5.4 kernel is broken for Pi 4 OTG and will not work in it's current state.

I've tried testing UVC gadget on Pi4 lately and hit issues here. For me it seems the whole pi4 gets hung. (Running on the default raspberry pi v5.4 kernel).

Is that the effect you saw? And has this been reported to RPi directly?
(

kbingham:
Aha,
> @richard-massless Setting kernel to 4.19 which is known to work with Pi4 OTG 

Ok - so this is good news, if we know that 4.19 works, we can bisect the kernel to find the issue.
Has anyone tested with a mainline kernel?

skarard:
> Ok - so this is good news, if we know that 4.19 works, we can bisect the kernel to find the issue.
> Has anyone tested with a mainline kernel?

https://bugs.launchpad.net/ubuntu/+source/linux-raspi2/+bug/1861070

dgsiegel:
This looks great!

@jeaton2411 would you be able to test on the Mac again?

@skarard / @richard-massless it would be great if we could use the latest kernel and we could reduce the kernel config a bit more. Other than that it looks good to me. It might also be worth to upgrade to `linux-image-5.4.0-1012-raspi` as described in the bug report above.

jeaton2411:
Happy to test & review, to be clear though, I have basically no skills beyond that (no git-fu or build skills). Point me at and image and I'll report back.

ainen:
> Happy to test & review, to be clear though, I have basically no skills beyond that (no git-fu or build skills). Point me at and image and I'll report back.

I've gone ahead and built the image myself from @skarard 's fork. It works well for me on Windows 10.

[sdcard-raspberrypi4-v1.60.img.gz](https://github.com/showmewebcam/showmewebcam/files/5661526/sdcard-raspberrypi4-v1.60.img.gz)

jeaton2411:
> > Happy to test & review, to be clear though, I have basically no skills beyond that (no git-fu or build skills). Point me at and image and I'll report back.
> 
> I've gone ahead and built the image myself from @skarard 's fork. It works well for me on Windows 10.
> 
> [sdcard-raspberrypi4-v1.60.img.gz](https://github.com/showmewebcam/showmewebcam/files/5661526/sdcard-raspberrypi4-v1.60.img.gz)

Testing this image on Macs. Seems to work about the same as last time, but with significantly less lag.  Still get horizontal scrambled lines of video in most apps: Facetime, Webcamera.io in Chrome, Screencastify in Chrome.  Seemingly randomly, Chrome based camera apps will return a usable 4x3 image letterboxed with a green bar on the side instead of a scrambled video feed.  Like last time, if I open PhotoBooth before or while I am getting either scrambled video OR the green letter box feed, the feed will fix itself to a full resolution 16x9 image but then lags badly.

If I hazard a guess, I'd say Mac apps are receiving different resolutions than they are requesting hence the scrambled image and weird green letter box.  Photo Booth seems to be overriding this with a max resolution request which solves the issue, but imparts lag as the full res image is hard for the web-based tools to handle.

Hope this information is helpful in the long run.

skarard:
The main issue with the builds based on the 4.19 kernel for the Raspberry Pi 4 is that you cannot change the video resolution. The video is fixed to 1600x1200. I haven't looked into why that is because I have been focusing on upgrading to the 5.4 kernel. 

I'm out of my depth when trying to debug kernel boot problems. I can't boot into a console and I think it may be todo with systemd not initializing. There are no obvious error messages when booting. The system hasn't hung as the boot screen accepts keystrokes and is responsive to USB devices.

My pi4 kernel 5.4 explorations are in: https://github.com/skarard/showmewebcam/tree/pi4-kernel-5.4

skarard:
After more testing, I am getting an RCU exception. It's hit or miss if it boots. I wonder if in this case, my machine is causing a problem.

@jeaton2411 Could you test this latest image on mac?
@ainen This version supports different resolutions.
@kbingham I have had this running on my pi4. Though every so often I start getting a `rcu_sched detected stalls on CPIs/tasks`. A reboot fixes it for me, my USB ports can be iffy sometimes so I don't know if it's just me. Could you let me know how it runs for you?

https://github.com/skarard/showmewebcam/releases/tag/V1.70-pi4-alpha

ainen:
Unfortunately this version is not detected at all on my Windows 10 machine. I tried building it myself and no luck.

jeaton2411:
Same experience here, neither image is detected as cam on Mac or PC.  Reverted to 1.6 to rule out hardware issues.  1.6 was detected.

skarard:
Hey Guys,

I tested my last couple of builds and the one that works for me is the same as the alpha but with console and kernel boot messages. To me, this suggests a race condition but I'm not sure where. @jeaton2411, @ainen, and @kbingham does this image boot for you guys too?

https://github.com/skarard/showmewebcam/releases/tag/V1.70-pi4-debug

The only difference between the debug version and the alpha is the debug enables a console on tty1 with `ln -sf /usr/lib/systemd/system/getty@.service "${TARGET_DIR}"/etc/systemd/system/getty.target.wants/getty@tty1.service` and disables the kernel boot messages by removing quiet from cmdline.txt.

jeaton2411:
1.7 debug version boots and mostly works for me on Mac and Win10.  Screencastify is struggling to open the stream on PC, but Webcamera.io and Zoom both seem ok with it on PC,  Seems to work with minimal lag in Mac as well.

ainen:
> I tested my last couple of builds and the one that works for me is the same as the alpha but with console and kernel boot messages. To me, this suggests a race condition but I'm not sure where. @jeaton2411, @ainen, and @kbingham does this image boot for you guys too?
> 
> https://github.com/skarard/showmewebcam/releases/tag/V1.70-pi4-debug
> 
> The only difference between the debug version and the alpha is the debug enables a console on tty1 with `ln -sf /usr/lib/systemd/system/getty@.service "${TARGET_DIR}"/etc/systemd/system/getty.target.wants/getty@tty1.service` and disables the kernel boot messages by removing quiet from cmdline.txt.

This version seems to work for me on Windows as well. However, the first boot didn't work. After a reboot it worked and I was able to serial in. The camera control utility is showing the resolution at 640x360 this time around and it is quite noticeable.

jeaton2411:
Screencastify just opened the stream successfully too, maybe unrelated issue causing trouble above.   The fixes to add back the resolutions seems to have made the Mac a lot happier.

parski:
Works great for me on an M1 Mac running Big Sur. Tried it in FaceTime, Photo Booth, OBS and Microsoft Teams.

configs/raspberrypi4
```diff
@@ -0,0 +1,8 @@
+BR2_arm1176jzf_s=n
+BR2_cortex_a72=y
+BR2_ARM_FPU_NEON_VFPV4=y
+BR2_LINUX_KERNEL_INTREE_DTS_NAME="bcm2711-rpi-4-b"
+BR2_PACKAGE_RPI_FIRMWARE_VARIANT_PI4=y
+BR2_KERNEL_HEADERS_5_4=n
+BR2_PACKAGE_HOST_LINUX_HEADERS_CUSTOM_5_4=y
```
dgsiegel, line 7: Why are you removing the 5.4 kernel headers, and then adding custom 5.4 headers below?

board/linux-raspberrypi4.config
```diff
@@ -0,0 +1,224 @@
...
+CONFIG_USB_F_RNDIS=m
+CONFIG_USB_F_MASS_STORAGE=m
+CONFIG_USB_F_FS=m
+CONFIG_USB_F_UAC1=m
+CONFIG_USB_F_UAC2=m
+CONFIG_USB_F_UVC=m
+CONFIG_USB_F_MIDI=m
+CONFIG_USB_F_HID=m
+CONFIG_USB_F_PRINTER=m
+CONFIG_BROKEN_ON_SMP=n
```
dgsiegel, line 16: Please check all =n settings if they are actually required. If not, let's remove them

configs/raspberrypi4
htruong: I think the PR author used buildroot 2020-3. It does add all the weird stuff to the defconfig.

htruong:
@skarard I *guess* you're using buildroot-2020.3 it looks like. Try using buildroot-2020.11 and lots of the differences will melt away.

jgrahamdev:
Testing latest build on a 2019 16" MBP, it works but poorly. Shortly after starting any app (including PhotoBooth) I start to see noise in the image. It progressively gets worse over time until the video feed freezes for that particular app. Restarting the app or opening another one appears to resolve the issue temporarily, but I once again get a ton of noise shortly after.

parski:
@skarard Is the plan to keep working on this on your [fork](https://github.com/skarard/pi4shadercam)? I'm still using the 1.70 debug build and it works great on Signal as well.

PR #52, opened 2020-11-27
reduce /boot partition size

dgsiegel:
this reduces the sdcard image by approx. 16mb


htruong:
Ah you forgot the pi 0w -- one more reason to support unified image:

```
 sdcard-raspberrypi0-3d76254.img 34.8 MB
 sdcard-raspberrypi0w-3d76254.img 50.8 MB
```

dgsiegel:
True. fixed.

Issue #56, opened 2020-11-29
Proposed Text Based Interface Control App


AlexOD42:
Attached is a proposed lightweight app for adjusting the webcam settings in the terminal before committing to /boot/camera.txt. The only dependency should be libcurses5. At present, the v4l2 control schema is hard-coded in the app based on my ZeroW and HD Sensor. I haven't handled the /boot filesystem remount or perhaps there's a better way? There may be a few bugs hiding in there but thought it was worth sharing for others to test. Please feel free to add to your repo if it's useful.

Thanks for all the great work on the wider project!

Alex

[control.c.txt](https://github.com/showmewebcam/showmewebcam/files/5612643/control.c.txt)
![screenshot](https://user-images.githubusercontent.com/74474237/100550577-d9ff9b80-3272-11eb-9820-0f21939a3b93.jpg)

htruong:
Thanks, let me add to new repo. I will credit you -- also do you want the source to be GPLv3 as well?

htruong:
The boot.txt remount is fine, it can be handled by a script, no big deal.

AlexOD42:
Perfect! Yes please to GPLv3.

dgsiegel:
Looks good! Agree that we should add a wrapper to handle camera.txt. Additionally, here's few thoughts:

- I think we should make use of a separate repo under the showmewebcam umbrella, as it would make it easier to integrate this tool via a buildroot package
- We would need a small Makefile, which handles compilation and installation of the binary (Here's an example you can use https://github.com/peterbay/uvc-gadget/blob/master/Makefile)
- Would it be possible to make the controls dynamic, ie. read them from v4l2? Here's some info on that: https://www.kernel.org/doc/html/v4.14/media/uapi/v4l/control.html
- You probably shouldn't use a hard coded filename in the code, make use of a parameter. That way you can also use it outside of showmewebcam
- Instead of using `system(cmd)` you want to use V4L2 controls like I mentioned above
- Do you already have a name for the app?

Love to include this for the next release, thanks for working on it!

peterbay:
I'm working on an implementation
![cam_control_001](https://user-images.githubusercontent.com/11430650/100587695-330e1480-32f1-11eb-977c-ca1aaa908f3f.png)

peterbay:
aplha version of camera-control - [https://github.com/peterbay/camera-control](https://github.com/peterbay/camera-control)

dgsiegel:
[...]

PR #57, opened 2020-12-01
add camera-control package

dgsiegel:
As discussed in #56


package/camera-control/camera-control.mk
```diff
@@ -0,0 +1,24 @@
...
+# camera-control
+#
+################################################################################
+
+CAMERA_CONTROL_VERSION = d7f5e46f5f0f9f53d3fece9f563ad5a019e9ca70
+CAMERA_CONTROL_SITE = git://github.com/peterbay/camera-control.git
+CAMERA_CONTROL_LICENSE = GPL-3.0+
+CAMERA_CONTROL_LICENSE_FILES = LICENSE
+CAMERA_CONTROL_DEST_DIR = /opt/camera-control
+CAMERA_CONTROL_SITE_METHOD = git
```
htruong, line 12: You need `CAMERA_CONTROL_DEPENDENCIES` to be `ncursesw` or something of that tune here, or this package will fail to build from time to time due to this package being built before `ncurses`.

package/camera-control/camera-ctl
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+mount -o remount,rw /boot || exit 1
+
+/opt/camera-control/camera-ctl -c /boot/camera.txt -v /dev/video0
```
htruong, line 5: This opens a bigger hole than needed for the boot to be vulnerable. Something like this will open the hole for it for only the bare minimum time needed...

```bash
#!/bin/sh

cp /boot/camera.txt /tmp/camera.txt || touch /tmp/camera.txt
/opt/camera-control/camera-ctl -c /tmp/camera.txt -v /dev/video0

mount -o remount,rw /boot || echo "Cannot remount boot as read-write, please inspect and transfer /tmp/camera.txt to /boot manually" && exit 1
cp /tmp/camera.txt /boot/camera.txt
sync
mount -o remount,ro /boot || echo "Warning: Cannot remount boot as read-only."
```

package/camera-control/camera-control.mk
dgsiegel: Thought the external packages will always be last, but good catch!

package/camera-control/camera-ctl
dgsiegel: Agree

dgsiegel:
Updated

htruong:
LGTM, feel free to merge whenerver you have a smoke test.

htruong:
Works fine for pi0w.

package/camera-control/camera-ctl
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
```
htruong, line 1: nit (you can change it or not): [The variables are not quoted](https://www.shellcheck.net/).

PR #58, opened 2020-12-02
Fix /boot parition size for pi0w

htruong:
The pi0w board is actually called pi0cam, so the PR#52 does not actually
change its boot partition size. Now fix the pi0w boot partition size for
real.


dgsiegel:
If that is the actual file, it is missing `camera.txt`as well.

Additionally we should get rid of all unused files

PR #62, opened 2020-12-06
improve build script to incluce targets more easily

dgsiegel:
this allows us to add different boards more easily, both
with buildroot as well as kernel configurations


htruong:
LGTM, thanks for doing this!

PR #64, opened 2020-12-07
Updated readme with Mac OS example

msurguy:
`ll` command doesn't exist on MacOS.

Also, I didn't need to run sudo to use `screen`, don't know if that's necessary for sure or not.


htruong:
Looks good to me, thanks for sending it in!

PR #66, opened 2020-12-08
bump buildroot and github runner versions to their latest stable

dgsiegel:
Let's see if the Github build succeeds now.


tuyenld:
@dgsiegel very nice.
I spent a whole last night to figure out what is wrong with Github Action.

tuyenld (commented):
```  
release:
    if: startsWith(github.ref, 'refs/tags/v')
    runs-on: ubuntu-latest
    needs: build
```
Do we need to change machine in the release action?

.github/workflows/release.yml
```diff
@@ -18,9 +18,9 @@ defaults:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
```
htruong, line 21: I like this. This way we don't have to worry about suddenly the build system breaks anymore :+1:

configs/config
```diff
@@ -1,7 +1,5 @@
 BR2_arm=y
 BR2_arm1176jzf_s=y
-BR2_CCACHE=y
```
htruong, line 3: Does this do or not do anything?

.github/workflows/release.yml
dgsiegel: > ```
> release:
>     if: startsWith(github.ref, 'refs/tags/v')
>     runs-on: ubuntu-latest
>     needs: build
> ```
> 
> Do we need to change machine in the release action?

Yes we do. Thanks for catching that.

configs/config
dgsiegel: ccache was the reason why the builds didn't work anymore, see my commit message:

    fix github actions build
    
    The build failed due to a conflict with ccache and cmake,
    which arose due to a recent version upgrade of cmake in the
    Github Ubuntu image.
    
    As ccache didn't give us any value on a cold cache, and
    little value on a hot cache (all buildroot packages are
    already built and we only need to compile changed packages)
    I see little value in maintaining it.
    
    Additionally this change bumps the Buildroot version to the
    latest stable version and uses ubuntu-20.04 which will
    become ubuntu-latest anyway soon.

htruong:
LGTM please feel free to merge whenever :)

PR #71, opened 2020-12-12
Turn off HDMI on startup

htruong:
Turning of HDMI saves 25mA of power


htruong:
Comparison

Before

IMG_4940

After

IMG_4941

configs/config
```diff
@@ -22,6 +22,7 @@ BR2_LINUX_KERNEL_NEEDS_HOST_OPENSSL=y
 BR2_LINUX_KERNEL_LZ4=y
 BR2_PACKAGE_RPI_FIRMWARE=y
 BR2_PACKAGE_RPI_FIRMWARE_X=y
+BR2_PACKAGE_RPI_USERLAND=y
```
dgsiegel, line 25: You're installing all of RPi userland (e.g. raspivid, raspistill, dtoverlay etc.) here. Is that what we want?

htruong: I kinda want to keep raspistill and raspivid to do debugging for end-users anyways. Those are nice. 

All the others can go but since we can't patch the makefile it'd be more pain than it's worth to not install them. What do you think?

htruong: Maybe TODO in another PR https://github.com/showmewebcam/showmewebcam/issues/75

dgsiegel: It's just that some userland tools break, as necessary libraries are not found.

htruong: OK, I will clean it up later in issue #75

htruong:
I think this is probably safe to merge. Should we do it @tuyenld and @dgsiegel ?

tuyenld:
@htruong 
I concern about "Preview Mode" on Pi camera. For example, `raspistill` displays image on HDMI even we are running Pi Lite version.
This mode virtually executes on GPU and it has a built-in "Overlay" feature.
https://picamera.readthedocs.io/en/release-1.13/recipes1.html?highlight=overlay#overlaying-images-on-the-preview 
If you turn it off, we may have problems to take advantage of the overlay feature.

dgsiegel:
> I think this is probably safe to merge. Should we do it @tuyenld and @dgsiegel ?

Not until we understood all implications of somewhat broken apps and libs with userland?

On another note, could you try if you can get the same result with executing `vcgencmd display_power 0`? That might help as well :)

PR #72, opened 2020-12-12
Test - Do not merge yet: Add 60fps support

### LINKED ISSUE #60: FPS Control
I'm using a Pi Zero (No W) with HQ Camera. Hypothetically, this device should support 720p60fps. How can I enable this setting?

htruong:
Fixes #60 (what a coincident)


wherget:
I have a conversion function fps to frameInterval lying around in my frame config branch https://github.com/showmewebcam/showmewebcam/compare/v1.80...wherget:fps-config
As far as I understood it, not all FPS will work with all resolutions (due to bandwith limitations), so it will probably need to be configured per-resolution whether 60fps is available.

I propose adding such a feature after #128 lands.

PR #76, opened 2020-12-14
Webui

tuyenld:
How to test
- Set host computer IP address: 10.0.0.2 netmask 255.255.255.252
- create `/boot/picam_web.txt`
- Open web browser: 10.0.0.1:8000


tuyenld:
We have live camera settings. 
WebUI look like this:

image

WebUI (front-end and back-end): https://github.com/tuyenld/picam-web

htruong:
This is awesome! Thanks for all the effort!

Let's try to merge your "expose more controls" first, and then my USB patch, and then let's try to merge this. 

One thing I kinda want to change is the way that piwebcam service is doing too many things now. I think it's probably time to separate control functions out of piwebcam and rename it to uvc-gadget before it becomes the monster that no one can touch.

At least I think the webui server should be its separate systemd unit. It can decide whether to startup depending the configuration.

peterbay:
Great, looks good, but -
1. Camera controls must be obtained from v4l2. Each camera module has implemented a different set of controls. 
2. Each control has min, max, default and actual value. Gui must have a possibility to reset single/all controls to the default value.
3. You can split the screen to the left and right parts. Left part with camera controls and right part with camera preview.
etc.. :-)

tuyenld:
@htruong 
> Let's try to merge your "expose more controls" first, and then my USB patch, and then let's try to merge this.

"expose more controls" commit: I still have no idea why some settings are missing on Linux (it works well on Windows).
However, I don't really like this idea. It completely depends on mapping/implementation between V4L2 and UVC.

@peterbay 
It just a prototype. Many works need to be done.

There is a similar project: https://github.com/silvanmelchior/RPi_Cam_Web_Interface
but I think it is too heavy on our system.

dgsiegel:
I'd be very vary to add too many features to showmewebcam. For example, if we manage to expose all settings to the host computer, any interface on the Pi itself would be an overload IMHO. In that case I'd probably create a native or web interface for the host computer instead of putting it onto the Pi.

package/piwebcam/multi-gadget.sh
```diff
@@ -1,30 +1,59 @@
...
+# {
+#     "name": "IBM Corporation",
+#     "field_vid": "1203"
+# },
+OS_CONFIG=$(echo $OS_CONFIG | tr '[A-Z]' '[a-z]')
+case $OS_CONFIG in
+  window)
+    # wismna hack for Window
+    # IBM Corporation RNDIS Driver will be loaded
+    echo 0x04b3 > idVendor
```
htruong, line 29: I don't think that we can pretend to be someone else and take their USB id.

package/piwebcam/multi-gadget.sh
```diff
@@ -1,30 +1,59 @@
...
-echo 0x1d6b > idVendor
-echo 0x0104 > idProduct
+
+# 0x04b3 = 1203
+# {
+#     "name": "IBM Corporation",
+#     "field_vid": "1203"
+# },
+OS_CONFIG=$(echo $OS_CONFIG | tr '[A-Z]' '[a-z]')
+case $OS_CONFIG in
```
htruong, line 25: Better way to do this is to have config file to see what usb gadget we want to configure, and the decision to ask what the pi wants to be is left to the user (we can suggest that if you use windows, then configure it as RNDIS, etc.)

htruong (commented):
Actually I am working on the PR to separate the configuration of the Pi as USB gadget. So stay tuned...

PR #78, opened 2020-12-19
piwebcam: Specify full target path to install command

antoneliasson:
This is needed to be able to build buildroot with
BR2_PER_PACKAGE_DIRECTORIES. Otherwise piwebcam.service would get
installed as the file /usr/lib/systemd/system rather than as a file
under /usr/lib/systemd/system/ since this directory did not already
exist when building with per-package directories.

Note the man page line about the -D option for install(1):

    create all leading components of DEST *except the last*


htruong:
Nice catch!

Issue #53, opened 2020-11-28
Only expose console login when a sentinel is present


htruong:
[Following the practice in raspbian](https://www.raspberrypi.org/documentation/remote-access/ssh/), we should consider only allowing ssh/console access if we see a certain sentinel, say `ssh`, placed in `/boot`.

dgsiegel:
@htruong can we close this now?

htruong:
Sure! Closing!

PR #79, opened 2020-12-20
Separate usb-gadget-config logic from uvc-webcam

htruong:
- Separate the logic to initialize usb gadget from uvc-webcam.
- Allow usb serial to be functional when uvc-webcam can't start up.
- Allow user to edit the usb-gadget.txt file to disable serial
access to the Pi.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 buildroot
 output
+/enable-serial-debug
 /camera.txt
 configs/*_defconfig
 board/linux.config
--- a/README.md
+++ b/README.md
@@ -82,9 +82,27 @@ $ screen /dev/tty.usbmodem13103 115200
 If the terminal is blank try pressing enter to see the login prompt. To exit
 the session, use `Ctrl-A \` (screen) or `Ctrl-A X` (minicom & picocom).
 
-**ATTENTION**: This interface is not really helpful if the Raspberry Pi didn't
-boot fully, as the serial-over-USB interface only comes up when uvc-gadget
-starts.
+**Warning**: This serial debug interface is automatically enabled and is controlled
+by the file called `enable-serial-debug` in the `/boot` folder. This is a potential
+security issue. For now, you should strongly consider disabling this feature by
+removing the file after you have finished customizing the webcam.
+
+
+### My camera doesn't show up on my host computer! What to do?
+
+From version 1.80, on Linux, you can see what happens by watching `dmesg`
+before you plug in the webcam:
+
+```
+$ sudo dmesg -w
+```
+
+If you only see the `ttyACM` device show up, but not the webcam, it's likely you
+have not plugged in the camera cable correctly, or the camera cable has gone bad.
+
+If you see nothing, maybe your USB cable is bad, or you have plugged in the cable
+to the wrong port.
+
 
 ## Customizing camera settings
 
--- a/board/genimage-raspberrypi0.cfg
+++ b/board/genimage-raspberrypi0.cfg
@@ -9,6 +9,7 @@ image boot.vfat {
       "rpi-firmware/start.elf",
       "rpi-firmware/overlays",
       "camera.txt",
+      "enable-serial-debug",
       "zImage"
     }
   }
--- a/board/genimage-raspberrypi0w.cfg
+++ b/board/genimage-raspberrypi0w.cfg
@@ -9,6 +9,7 @@ image boot.vfat {
       "rpi-firmware/start.elf",
       "rpi-firmware/overlays",
       "camera.txt",
+      "enable-serial-debug",
       "zImage"
     }
   }
--- a/board/post-image.sh
+++ b/board/post-image.sh
@@ -90,6 +90,17 @@ __EOF__
 			cat "$BR2_EXTERNAL_PICAM_PATH/camera.txt" >> "${BINARIES_DIR}/camera.txt"
 		fi
 
+		# Add default enable-serial-debug file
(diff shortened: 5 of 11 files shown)


htruong:
- Tested this on my pi0w with and without the camera connected.

- Also tested with ECM+Camera+Serial on Linux. (ECM doesn't actually do anything, you can only ping the pi for now.)

dgsiegel:
Here's a few thoughts on this approach:

- I don't think we should add a second BR package, piwebcam is fine as the functionality is almost the same and it doesn't give us any value to maintain a different package, when the shell and systemd scripts are almost the same. Also the `src` thing doesn't look right to me
- Can we move network/RNDIS/ECM to a different PR? It looks like this could break quite easily (e.g. the static IP address which I commented on already in a different PR, the MAC address and so on)
- I don't think we should allow configuration of the usb gadget. Most people download the image anyway, and just want to fire & forget. Webcam and Serial should always work.

htruong:
All points addressed.

>I don't think we should add a second BR package

Yeah brought back the `multi-gadget.sh` script to piwebcam. The `src` folder was a hack, it is so that buildroot doesn't complain about the `_SITE` ends up in a slash.

>Can we move network/RNDIS/ECM to a different PR?

Moved.

>I don't think we should allow configuration of the usb gadget. Most people download the image anyway, and just want to fire & forget. Webcam and Serial should always work.

You can imagine someone would want to not expose the serial interface for security reasons. I kinda regret allowing the debug interface by default, especially when the username and password is root/root and cannot be changed. It can be hacked and persistently owned.

dgsiegel:
> You can imagine someone would want to not expose the serial interface for security reasons.

Well, it is only connected via USB and you need root anyway to access the serial console (on Linux anyways).

I kinda regret allowing the debug interface by default, especially when the username and password is root/root and cannot be changed (this is true for most normal users). It can be hacked and persistently owned if any computer it plugged to is compromised. Then it can be used to change the showmewebcam image and turn it into a virus vector to other computers (aka stuxnet).

Not really feasible, as we use SquashFS and an upgrade currently wipes all changes on the SD Card. In any case, if you really really don't want to expose the debug interface by default, I'd rather use a single file, e.g. `/boot/console`, instead of a config file, as proposed in #53.

My major issue with the config file right now is that `CONFIGURE_USB_WEBCAM` is useless and `CONFIGURE_USB_SERIAL` could be exposed in a simpler fashion.

htruong:
>Not really feasible, as we use SquashFS and an upgrade currently wipes all changes on the SD Card.

If you keep using one version that is. When you upgrade it's already too late, the virus has spread and persisted by other means.

SquashFS only makes it inconvenient to write to the file system, it doesn't make it impossible. One way to achieve persistence ownage is write another image right after the SquashFS then alter the partition table. Another way to do it is to copy the rootfs to `/tmp` and switch root, then you can do whatever you want with the SquashFS partition on the SD card. Another way to achieve it is to just write directly on top of the old SquashFS partition and the user would be none the wiser (even when the user is actively using the webcam -- our rootfs is so small that everything seems to be cached in RAM, you can check that by pulling the SD Card while the webcam is running).

All those things are doable just because we don't protect the firmware root access by default.

>My major issue with the config file right now is that CONFIGURE_USB_WEBCAM is useless and CONFIGURE_USB_SERIAL could be exposed in a simpler fashion.

Regarding point 2 I think the idea is to get the network config in, so really the webcam is needed. Having one config file instead of having sentinels all over the places to me is more organized.

donbindner:
The serial interface is really great for configuring the camera, and it "just works."  But it really is insecure.

If the v4l interface is capable enough, the default release should probably not have serial.  If someone is having trouble or is willing to understand even just a little bit, there should be a debug build.

Having a default password is so obviously wrong that it's going to get someone in trouble in a way we can't even imagine yet.

dgsiegel:
> All those things are doable just because we don't protect the firmware root access by default.

True. So how about adding a user and disabling root access?

> If the v4l interface is capable enough, the default release should probably not have serial. If someone is having trouble or is willing to understand even just a little bit, there should be a debug build.

Right, so in the end we could remove the serial interface once we can expose all v4l settings to the host?

Adding the serial interface for debug builds could just be a flag in our build script.

> Having a default password is so obviously wrong that it's going to get someone in trouble in a way we can't even imagine yet.

Agree.

htruong:
I would like to get this in, because this is a step in the right direction that will allow us to have more flexibility to do things. The serial disable switch will get us to where we want to be once we can figure out the automatic save for camera settings.

dgsiegel:
> I would like to get this in, because this is a step in the right direction that will allow us to have more flexibility to do things. The serial disable switch will get us to where we want to be once we can figure out the automatic save for camera settings.

Okay, but can we at least agree to not have a config file there? :) Let's just check if a file exists in `/boot`?

htruong:
> Okay, but can we at least agree to not have a config file there? :) Let's just check if a file exists in `/boot`?

Sure, done that with the latest push.

htruong:
Tested on my pi0w, serial working by default, but no more serial after you delete the sentinel file.

dgsiegel:
Last issue: I don't really like to have `enable-serial-debug` in the root folder. Can we either create it dynamically in `post-image.sh` or use the overlay to copy the file?

htruong:
Done!

dgsiegel:
You might want to revert .gitignore, other than that this looks good to me.

dgsiegel (approved):

htruong:
Will do it in the fixes PR

PR #81, opened 2020-12-20
fix github actions cache key

dgsiegel:
it seems like that secrets was not available on a global env level, but only during the steps


htruong (approved):
Thanks for doing this!

PR #82, opened 2020-12-22
Small fixes and improvements

htruong:
Some small fixes:

- Do not generate the ext2fs image while building, pretty useless
- Move the rootfs to rootfs on repo root, so it's more discoverable


tuyenld:
I sometime use `ext4` to generate debugging image, I can change everything in rootfs
I think we need to keep it for debugging too.

dgsiegel:
> I sometime use `ext4` to generate debugging image, I can change everything in rootfs
> I think we need to keep it for debugging too.

If you just want to add new files, you can use the overlay. If you actually want to change files, a debug flag might make sense, as we probably will need that for the serial interface too at some point

@htruong would you mind using `.gitkeep` instead of `placeholder`? It seems that this file gets copied over, while all dot files shouldn't, if I understood the docs correctly.

htruong:
Fixed. 

Yeah, I think we can have a debug build that will have ext4 image, that's going to be very nice. However, I think for now we should remove this because it does nothing.

dgsiegel (approved):
Looks good to me

htruong:
Tested on my pi0w, works well enough for me!

PR #83, opened 2020-12-24
Github actions: Cache buildroot

htruong:
Currently, every time we build, we hammer the server of our
good friends at Buildroot project. We should not do that, and cache
our extracted tarball.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -42,19 +42,23 @@ jobs:
           echo "GIT_TAG_NAME=${GITHUB_REF#refs/tags/}" >> $GITHUB_ENV
           echo "Current version: $(git describe)"
 
+      - name: Cache Build
+        id: restore-cache
+        uses: actions/cache@v2
+        with:
+          path: |
+            output
+            buildroot
+          # Stupid hack to invalidate cache from UI
+          # https://github.com/actions/cache/issues/2#issuecomment-673493515
+          key: ${{ matrix.raspberry }}-${{ env.BUILDROOT_DIR }}-${{ env.BUILDROOT_VERSION }}-${{ secrets.CACHE_VERSION }}
+
       - name: Download Buildroot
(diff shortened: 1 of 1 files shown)


htruong:
Save a precious time of 2 seconds for us and 6.5 MB for our friends each PR!

.github/workflows/release.yml
```diff
@@ -42,13 +42,24 @@ jobs:
...
         run: |
           wget -qc https://buildroot.org/downloads/buildroot-${{ env.BUILDROOT_VERSION }}.tar.gz
           tar -zxf buildroot-${{ env.BUILDROOT_VERSION }}.tar.gz
           mv buildroot-${{ env.BUILDROOT_VERSION }} ${{ env.BUILDROOT_DIR }}
 
       - name: Cache Build
+        id: restore-cache-output
         uses: actions/cache@v2
         with:
           path: output
```
dgsiegel, line 65: it would be easier to just add the buildroot path here :) that way we don't have to work with two cache files

.github/workflows/release.yml
```diff
@@ -42,13 +42,24 @@ jobs:
...
+      - name: Cache Buildroot
+        id: restore-cache-br
+        uses: actions/cache@v2
+        with:
+          path: buildroot
+          # Stupid hack to invalidate cache from UI
+          # https://github.com/actions/cache/issues/2#issuecomment-673493515
+          key: br-${{ env.BUILDROOT_DIR }}-${{ env.BUILDROOT_VERSION }}-${{ secrets.BR_CACHE_VERSION }}
+
       - name: Download Buildroot
```
dgsiegel, line 54: this step would then have to go after the cache checkout

.github/workflows/release.yml
htruong: Good point, trying again with your suggestion!

dgsiegel (approved):
Looks good!

PR #84, opened 2020-12-24
Github actions: Run shellcheck

htruong:
- Run shellcheck on PRs and releases to prevent silly shell mistakes

- Fix post-image shellcheck warning

### DIFF
--- a/.github/workflows/shellcheck.yml
+++ b/.github/workflows/shellcheck.yml
@@ -1,0 +1,27 @@
+name: Shellcheck
+
+on:
+  # Triggers the workflow on push or pull request events on the master branch
+  pull_request:
+    branches: [ master ]
+
+  # Allows to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  shellcheck:
+    runs-on: ubuntu-20.04
+
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: Shellcheck all scripts
+        run: |
+          shellcheck $(git ls-files '*.sh')
--- a/board/post-image.sh
+++ b/board/post-image.sh
@@ -2,7 +2,7 @@
 
 set -e
 
-GENIMAGE_CFG="$(dirname $0)/genimage-${BOARDNAME}.cfg"
+GENIMAGE_CFG="$(dirname "$0")/genimage-${BOARDNAME}.cfg"
 GENIMAGE_TMP="${BUILD_DIR}/genimage.tmp"
 
 for arg in "$@"


htruong:
Example of failed shellcheck PR: https://github.com/htruong/showmewebcam/runs/1603732887?check_suite_focus=true

shellcheck-all.sh
```diff
@@ -0,0 +1,24 @@
...
+
+TMP_DIR=smw-build-tmp
+SCRIPTS_FILE="$TMP_DIR"/bash-scripts.txt
+ALL_PASSED_SC=true
+
+mkdir -p "$TMP_DIR"
+
+find . -not -path './buildroot/*' -name '*.sh' -type f > "$SCRIPTS_FILE"
+
+while read -r s; do
```
dgsiegel, line 11: It would be better to use find directly in the loop or to use globbing to get all sh files

shellcheck-all.sh
```diff
@@ -0,0 +1,24 @@
...
+  echo "Checking $s..."
+  if ! ./shellcheck/shellcheck "$s" ; then
+    echo "Error: File $s failed shellcheck."
+    ALL_PASSED_SC=FALSE
+  fi
+done < "$SCRIPTS_FILE"
+
+if [ "$ALL_PASSED_SC" = true ] ; then
+  echo "Congratulations, your scripts look good."
```
dgsiegel, line 20: No need to add output here?

shellcheck-all.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
```
dgsiegel, line 1: Do you we need this file outside of github actions? If not, the content could easily go into the actions file, which supports shell commands too.

shellcheck-all.sh
```diff
@@ -0,0 +1,24 @@
...
+
+TMP_DIR=smw-build-tmp
+SCRIPTS_FILE="$TMP_DIR"/bash-scripts.txt
+ALL_PASSED_SC=true
+
+mkdir -p "$TMP_DIR"
```
dgsiegel, line 7: use `mktemp`

.github/workflows/release.yml
```diff
@@ -17,6 +17,47 @@ defaults:
     shell: bash
 
 jobs:
+  shellcheck:
```
dgsiegel, line 20: it might make sense to move this to a different github actions script which only runs on pull requests

.github/workflows/release.yml
```diff
@@ -17,6 +17,47 @@ defaults:
...
+
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      # Retrieve annotated tags as the checkout above doesn't do this
+      # https://github.com/actions/checkout/issues/217#issuecomment-636481619
+      - name: Fetch Git Tag History
```
dgsiegel, line 34: we don't need the full history here, we just need the current state of files

.github/workflows/release.yml
```diff
@@ -17,6 +17,47 @@ defaults:
...
+          echo "Current version: $(git describe)"
+
+      - name: Cache Shellcheck
+        id: restore-cache-sc
+        uses: actions/cache@v2
+        with:
+          path: shellcheck
+          # Stupid hack to invalidate cache from UI
+          # https://github.com/actions/cache/issues/2#issuecomment-673493515
+          key: sc-${{ env.SHELLCHECK_DIR }}-${{ env.SHELLCHECK_VERSION }}-${{ secrets.SC_CACHE_VERSION }}
```
dgsiegel, line 48: no need for caching here, we're just checking out the current git repo

.github/workflows/release.yml
```diff
@@ -17,6 +17,47 @@ defaults:
...
+
+      - name: Download Shellcheck
+        if: steps.restore-cache-sc.outputs.cache-hit != 'true'
+        run: |
+          wget -qc https://github.com/koalaman/shellcheck/releases/download/v${{ env.SHELLCHECK_VERSION }}/shellcheck-v${{ env.SHELLCHECK_VERSION }}.linux.x86_64.tar.xz
```
dgsiegel, line 53: why not use `apt-get install shellcheck -y`?

shellcheck-all.sh
htruong: If you use find and shellcheck individually, then the script will fail and exit on the first file with shell errors/warnings. So a person modifying or adding a whole bunch of (failing) scripts that doesn't have shellcheck installed will have to fix it a couple of times before having all issues in their script solved. That is because each time they only know that only one additional file failed. Here if we check all and then fail then they will have a list of files that failed, so it saves them some time.

shellcheck-all.sh
htruong: Agreed.

shellcheck-all.sh
htruong: Probably, that will allow the PR submitter to run this script on their dev machine before having to push to see the results. That way they don't have to dig into the yml code to see what they have to do in order to pass shellcheck.

.github/workflows/release.yml
htruong: Probably, I guess I don't want to hit Ubuntu's servers if we can use our caching to restore the installation. But it's probably nothing to them and/or github has their own Ubuntu repo so they don't hit Ubuntu each time the developers do `apt-get`.

dgsiegel: No worries, they've mirrored all packages locally. Feel free to use apt-get here.

shellcheck-all.sh
dgsiegel: That depends. If you're just calling shell-check for a single file, then the script doesn't bring any additional value. They will need to install shell-check anyways (or at least use the online version).

shellcheck-all.sh
dgsiegel: Actually we don't need a loop here. shellcheck can have multiple arguments, e.g.

`shellcheck $(git ls-files '*.sh')

Alternatively you can also add flags to ls-files or just use find, ignoring output and buildroot.

.gitignore
```diff
@@ -1,3 +1,4 @@
+smw-build-tmp
```
dgsiegel, line 1: no need for this anymore

htruong:
Yep, the new PR is much cleaner and the effect is still the same.

Sample of failed Shellcheck PR: https://github.com/htruong/showmewebcam/runs/1612768249?check_suite_focus=true

htruong:
@dgsiegel would appreciate another review whenever you have time :)

dgsiegel:
Looks good! Just wondering whether the action fails if shellcheck finds something?

htruong:
Absolutely. It will output something like this. https://github.com/htruong/showmewebcam/runs/1612768249?check_suite_focus=true

dgsiegel:
Awesome

dgsiegel (approved):

PR #93, opened 2021-01-27
Enable additional UVC control settings

dognotdog:
This PR adds a patch file that enables all UVC video settings in the USB descriptor for which V4L2 mappings exist in `uvc-gadget`, though not all settings are fully functional, yet (neither are all settings functional with the 0001 patch).

Camera Terminal in uvc-gadget maps 0x0E / 14, 0x00, 0x00
D1 AE_MODE
D2 AE_PRIORITY
D3 exposure time

Processing Unit in uvc-gadget maps 0xFF / 255,  0x2F / 47, 0x00
D0 Brightness
D1 contrast
D2 hue
D3 saturation
D4 sharpness
D5 gamma
D6 WB temperature
D7 WB component

D8 Backlight compensation
D9 gain
D10 flicker
D11 auto hue
D13 WB component auto maps to V4L2 auto WB

### DIFF
--- a/patches/linux-custom/0002-linux-enable-even-more-video-controls.patch
+++ b/patches/linux-custom/0002-linux-enable-even-more-video-controls.patch
@@ -1,0 +1,22 @@
+diff --git a/drivers/usb/gadget/function/f_uvc.c b/drivers/usb/gadget/function/f_uvc.c
+index 5c850be..5a9ec10 100644
+--- a/drivers/usb/gadget/function/f_uvc.c
++++ b/drivers/usb/gadget/function/f_uvc.c
+@@ -802,7 +802,7 @@ static struct usb_function_instance *uvc_alloc_inst(void)
+        cd->wObjectiveFocalLengthMax    = cpu_to_le16(0);
+        cd->wOcularFocalLength          = cpu_to_le16(0);
+        cd->bControlSize                = 3;
+-       cd->bmControls[0]               = 2;
++       cd->bmControls[0]               = 14;
+        cd->bmControls[1]               = 0;
+        cd->bmControls[2]               = 0;
+ 
+@@ -815,7 +815,7 @@ static struct usb_function_instance *uvc_alloc_inst(void)
+        pd->wMaxMultiplier              = cpu_to_le16(16*1024);
+        pd->bControlSize                = 2;
+        pd->bmControls[0]               = 255;
+-       pd->bmControls[1]               = 6;
++       pd->bmControls[1]               = 47;
+        pd->iProcessing                 = 0;
+ 
+        od = &opts->uvc_output_terminal;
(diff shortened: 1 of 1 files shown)


dgsiegel:
It seems that not all settings are exposed to the host?

$ v4l2-ctl -d /dev/video2 -l
                     brightness 0x00980900 (int)    : min=0 max=100 step=1 default=50 value=50
                       contrast 0x00980901 (int)    : min=0 max=200 step=1 default=100 value=98
                     saturation 0x00980902 (int)    : min=0 max=200 step=1 default=100 value=105

Also, which ones are functional and which ones aren't?

dognotdog:
The settings in black work, the ones that are grayed out are mapped in `uvc-gadget`, but don't seem to work because `GET_INFO` returns `0x00` instead of the right flags. It seems like there is an `enabled` flag in the mapping, but it is not getting enabled because the [ioctl() call](https://github.com/peterbay/uvc-gadget/blob/0054986bb3d3a814fec9a955a4b11f3903da0680/uvc-gadget.c#L975) that checks availability on `v4l` doesn't see it -- it might not be available, though why exactly I am unsure, that is as far as I've gotten in the chase.

image

htruong:
LGTM. Any opposition, @dgsiegel ?

dgsiegel:
> LGTM. Any opposition, @dgsiegel ?

Well I couldn't get it to work on my machine, see above.

dognotdog:
> > LGTM. Any opposition, @dgsiegel ?
> 
> Well I couldn't get it to work on my machine, see above.

This patch only touches the UVC `bmControl` bits in the descriptor, not what v4l actually does — some are extended controls, which are not listed by 
`$ v4l2-ctl -d /dev/video2 -l`
 because that only lists `VIDIOC_QUERYCTRL` and not `VIDIOC_QUERY_EXT_CTRL` controls.

htruong:
Yeah I can confirm @dgsiegel 's finding that my Linux host only shows three controls. Sorry for the confusion, yeah this is not working for Linux at least.

```
$ v4l2-ctl -d /dev/video4 -l
                     brightness 0x00980900 (int)    : min=0 max=100 step=1 default=50 value=50
                       contrast 0x00980901 (int)    : min=0 max=200 step=1 default=100 value=98
                     saturation 0x00980902 (int)    : min=0 max=200 step=1 default=100 value=105
```

dognotdog:
@htruong could you check if the `bmControl` fields in the descriptor are as expected? Also, `-l` only lists `VIDIOC_QUERYCTRL` controls, according to the man page, which are only a subset of the possible UVC settings.

dgsiegel:
I guess this is almost the same as #118, which allows to configure the bits on runtime. Ok if we close this in favor of it?

dgsiegel:
#118 was merged.

PR #100, opened 2021-02-08
A camera control script that automates running camera-ctl 

akseidel:
This is a shell script with accompanying instruction readme that performs all tasks required to start camera-ctl in a Terminal session.

This script when executed in a Terminal window finishes up with "camera-ctl" running in a screen session. It automates the serial connection after detecting the tty serial device presence, logs in as root and runs "camera-ctl".

Although the showmewebcam readme describes "camera-ctl" in debugging it is actually an essential utility required to be running when showmewebcam is being used in dynamic situations. This script allows ordinary users to easily invoke "camera-ctl".

### DIFF
--- a/picamctl/picamctl.sh
+++ b/picamctl/picamctl.sh
@@ -1,0 +1,148 @@
+#!/bin/bash
+# akseidel 02/06/2021
+# A script automating the steps to run showmewencam's "camera-ctl".
+
+# The script will pause for continue or exit if run with an argument.
+debugpause=$1 
+# edit this next line to be the correct text pattern for your computer.
+portnamepat="tty.usbmodem*" #typical for showmewebcam on Appl OS X
+#portnamepat="ttyACM*" #probably typical for showmewebcam on Linux
+#portnamepat="tty.*"
+# The client is an application you want this script to also run. 
+runtheclient="true"
+clientname="Photo Booth"
+
+# initial cleanup
+initclean(){
+    reset
+    clear
+    # terminates all the ort screen sessions.
+    screen -ls | grep Attached | cut -d. -f1 | awk '{print $1}' | xargs kill
+    screen -ls | grep Detached | cut -d. -f1 | awk '{print $1}' | xargs kill
+}
+
+# run the client application
+runclient(){
+    if [ "$runtheclient" = "true" ]; then
+        # -g causes application to open in background 
+        # https://scriptingosx.com/2017/02/the-macos-open-command/
+        open -g -a "$clientname"
+    fi
+}
+
+# start the screen session
+doscreensession(){
+    echo ""
+    echo "Now establishing serial connection using 'screen' ..."
+    # For some reason screen stuffing the user, password and camera-ctl to
+    # a detached target screen does not work without the target screen having
+    # been attached at some point. Here the nest screen into a spawner screen
+    # trick is used to get around that.   
+    screen -dmS spawner
+    screen -S spawner -X screen screen -dR thispicam $piusbwebcamport 115200
+    sleep 0.3
+    screen -S thispicam -X detach
+    sleep 0.3
+    screen -S thispicam -X stuff $'root\n'
+    sleep 0.3
+    screen -S thispicam -X stuff $'root\n'
+    sleep 0.3
+    screen -S thispicam -X stuff $'camera-ctl\n'
+    screen -r thispicam
+}
+
+# show intro text
+showintro(){
+    echo "======================================================================"   
+    echo "                             picamctl                                 "
+    echo "======================================================================"       
+    echo "  This script looks for a serial port device that could be the        "  
+    echo "  Showmewebcam USB webcam device. It will connect to the device, log  "
+    echo "  in as root and then run the camera settings control utility.        " 
+    echo "----------------------------------------------------------------------"
+    echo "  Looking for a device like: /dev/$portnamepat                        "
+}
+
+# collect the port names
+getportnames(){
+    serialportlist=$(ls -a /dev/$portnamepat 2> /dev/null)
+    countofserialportslist=$(ls -a /dev/$portnamepat  2> /dev/null | wc -l)
+}
+
+# report port qty
+rptportqty(){
+    echo "  $(expr $countofserialportslist + 0) such ports found."
+}
+
+# report ports
+rptportlist(){
+    echo " " $serialportlist 2> /dev/null
+}
+
+# set port to use and show it
+dosetport(){
+    # changes item space separator to a newline and returns tail item
+    piusbwebcamport=$(echo "$serialportlist" | tr ' ' '\n' | tail -1)
+    echo "  Assuming this last one => " $piusbwebcamport
+    echo "======================================================================"
+}
+
+# option for pause and exit if any argument present, for diagnostic use
+option2exit(){
+    if [ ! -z "$1" ]
+    then
+        while read -r -s -p "Press any key when ready to continue. (ESC key quits)" -n1 key 
+        do
+            # check for escape key
+            if [[ $key == $'\e' ]]; then
+                echo ""
+                echo "Ok, Bye"
+                exit 0
+            else
+                break
+            fi
+        done
+    fi
+}
+
+# checkboot status
+checkbootstatus(){
+while [ $countofserialportslist = 0 ]
+    do
+        echo ""
+        echo "  None found! Got that USB thing plugged in or waited the 10 seconds"
+        echo "  needed for booting up? Periodic boot checking is now happening."
+        echo "======================================================================"
+        echo ""
+        chk=1
+        chklim=12
+        while [ $chk -le $chklim ]
+            do
+                echo -ne "  Checking for a ready showmewebcam. Attempt $chk out of $chklim  ... \r"
+                getportnames
+                if [ $countofserialportslist != 0 ]; then
+                    break 2
+                fi
+                chk=$(( $chk + 1 ))
+                sleep 2
+            done
+        echo -e "  Run this again when the camera is ready.                             "
+        echo ""
+        exit 
+    done
+}
+
+# program section
+
+initclean
+showintro
+getportnames
+rptportqty   
+checkbootstatus
+rptportlist
+dosetport
+option2exit $1
+runclient
+doscreensession
+
+# end        
--- a/picamctl/picamctl_README.md
+++ b/picamctl/picamctl_README.md
@@ -1,0 +1,41 @@
+# ScriptsForShowMeWebCam
+
+## picamctl.sh
+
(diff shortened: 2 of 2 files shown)


htruong:
Thanks for sending in the patch! It seems like the shellcheck did not pass, do you mind revising so that it passes Shellcheck?

picamctl/picamctl.sh
```diff
@@ -0,0 +1,146 @@
+#!/bin/bash
+# akseidel 02/06/2021
+# A script automating the steps to run showmewencam's "camera-ctl".
+# The script will pause for continue or exit if run with an argument.
+
+# edit this next line to be the correct text pattern for your computer.
+portnamepat="tty.usbmodem" #typical for showmewebcam on Appl OS X
+#portnamepat="ttyACM" #probably typical for showmewebcam on Linux
+#portnamepat="tty."
```
dgsiegel, line 9: You should auto discover this, based on the platform you are on.

picamctl/picamctl.sh
```diff
@@ -0,0 +1,146 @@
...
+runtheclient="true"
+clientname="Photo Booth"
+
+# initial cleanup
+initclean(){
+    reset
+    clear
+    # terminates all the ort screen sessions.
+    screen -ls | grep Attached | cut -d. -f1 | awk '{print $1}' | xargs kill
+    screen -ls | grep Detached | cut -d. -f1 | awk '{print $1}' | xargs kill
```
dgsiegel, line 20: Why two times?

picamctl/picamctl.sh
```diff
@@ -0,0 +1,146 @@
...
+# A script automating the steps to run showmewencam's "camera-ctl".
+# The script will pause for continue or exit if run with an argument.
+
+# edit this next line to be the correct text pattern for your computer.
+portnamepat="tty.usbmodem" #typical for showmewebcam on Appl OS X
+#portnamepat="ttyACM" #probably typical for showmewebcam on Linux
+#portnamepat="tty."
+# The client is an application you want this script to also run. 
+runtheclient="true"
+clientname="Photo Booth"
```
dgsiegel, line 12: MacOs specific, should be based on the platform, a parameter or optional

picamctl/picamctl.sh
```diff
@@ -0,0 +1,146 @@
...
+# akseidel 02/06/2021
+# A script automating the steps to run showmewencam's "camera-ctl".
+# The script will pause for continue or exit if run with an argument.
+
+# edit this next line to be the correct text pattern for your computer.
+portnamepat="tty.usbmodem" #typical for showmewebcam on Appl OS X
+#portnamepat="ttyACM" #probably typical for showmewebcam on Linux
+#portnamepat="tty."
+# The client is an application you want this script to also run. 
+runtheclient="true"
```
dgsiegel, line 11: Better as a parameter

picamctl/picamctl.sh
```diff
@@ -0,0 +1,146 @@
+#!/bin/bash
```
dgsiegel, line 1: #!/bin/sh or #!/usr/bin/env bash are better for platform compatibility.

picamctl/picamctl.sh
```diff
@@ -0,0 +1,146 @@
...
+    screen -ls | grep Attached | cut -d. -f1 | awk '{print $1}' | xargs kill
+    screen -ls | grep Detached | cut -d. -f1 | awk '{print $1}' | xargs kill
+}
+
+# run the client application
+runclient(){
+    if [ "$runtheclient" = "true" ]; then
+        # -g causes application to open in background 
+        # https://scriptingosx.com/2017/02/the-macos-open-command/
+        open -g -a "$clientname"
```
dgsiegel, line 28: On Linux, this would probably be xdg-open

picamctl/picamctl.sh
```diff
@@ -0,0 +1,146 @@
...
+doscreensession(){
+    echo ""
+    echo "Now establishing serial connection using 'screen' ..."
+    # For some reason screen stuffing the user, password and camera-ctl to
+    # a detached target screen does not work without the target screen having
+    # been attached at some point. Here the nest screen into a spawner screen
+    # trick is used to get around that.   
+    screen -dmS spawner
+    screen -S spawner -X screen screen -dR thispicam "$piusbwebcamport" 115200
+    sleep 0.3
```
dgsiegel, line 42: Why the delays? You can just read the output and wait for that.

picamctl/picamctl.sh
```diff
@@ -0,0 +1,146 @@
...
+
+# start the screen session
+doscreensession(){
+    echo ""
+    echo "Now establishing serial connection using 'screen' ..."
+    # For some reason screen stuffing the user, password and camera-ctl to
+    # a detached target screen does not work without the target screen having
+    # been attached at some point. Here the nest screen into a spawner screen
+    # trick is used to get around that.   
+    screen -dmS spawner
```
dgsiegel, line 40: One might need sudo to access the device with screen.

picamctl/picamctl.sh
akseidel: Ok, I will work on that. I have a Raspberry Pi 4 I can use for testing the Linux side.

picamctl/picamctl.sh
akseidel: This is my first shell script, so I have to dive deep to understand even the smallest aspects. Based on tests, what I had for what is supposed to handle both Attached and Detached in one line did not work. The single lines did. So I used those and moved on. There are many brute force parts of this script that I assume must have much better solutions.

picamctl/picamctl.sh
akseidel: Yes it is MacOS specific. The intent is for educator to make edits, which is one reason these are brought to the script top. The intended users (school setting as webcam attached to microscope camera attachment) would not know what a parameter is. I understand the point. Perhaps there is something that could be done.

picamctl/picamctl.sh
akseidel: In the intended situation running the client is almost 100% wanted and in MacOS it would be difficult to have multiple instances. This variable is almost an ort from the early development process. An issue with parameters is that they result in what is viewed as unnecessary complication by the intended user.

picamctl/picamctl.sh
akseidel: Ok, I will check this out.

picamctl/picamctl.sh
akseidel: Thanks. Hopefully I'll get the platform checking to work.

picamctl/picamctl.sh
akseidel: I don't know how to read the output. I hardly understand screens. Can you provide a reference link to start me on the right track to read screen output? The sleeps were necessary. Sometimes the Pi Zero was not ready in time. I thought it was lucky sleep would take decimal numbers instead of only integers. Perhaps by reading the screen output it might also be possible to detect the unrefreshed state camera-ctl can be in when the Pi Zero already had it running.

picamctl/picamctl.sh
akseidel: I am not sure what you mean. I can tell you it took a lot of testing to determine what was going on. Apparently one cannot stuff to a detached screen that has not been attached yet. Looking online knowing that observed behaviour led to this solution. The visible, ie attached, initial screen session blocks the script until that screen is terminated. Therefore the succeeding stuffs to provide root, root, camera-ctl do not occur until the first screen is terminated. Starting the initial screen detached allowed the succeeding code to proceed, but the stuffs vanish.

htruong:
Yeah I agree with @dgsiegel. 

I also understand the intention of @akseidel though, as you seem to have a pretty nice set up with your microscope. Perhaps having this example script in a separate sample directory for specific example use-cases would be OK for me too.

dgsiegel:
> Perhaps having this example script in a separate sample directory for specific example use-cases would be OK for me too.

Sounds good. We could put the swmc-expect script there as well.

picamctl/picamctl.sh
dgsiegel: Depending on the access mode of the device, you need sudo to access it.

akseidel:
I figured you all would decide how this script would fit in the repository, if it could fit in. I did not see a logical spot in the current repository. That is why I isolated it into its own folder.

I gather you might have seen the youtube video made regarding a microscope setup. I put that out soon after first getting it to work because of its potential in school settings. The setup is created for my wife’s school science department where she and other teachers, due to the pandemic, are trying to figure out ways to fit their lab course work into remote learning. This showmewebcam used this way is a game changer for them. For example it has been used to make videos of Daphina heart beats for students at home to figure out the beat rate. It has been used to make still images of anesthetized fruit flys for students at home to count genetic ratios in crossed fruit fly types. It provides a remote learning solution for all classroom lab sessions that involve using a microscope. We can not thank you enough for making this possible.

picamctl/picamctl.sh
akseidel: Auto discover is now implemented.

picamctl/picamctl.sh
akseidel: In combination with the auto discover implementation the script sets these per OS.

picamctl/picamctl.sh
akseidel: The script has been entirely refactored for sh.

picamctl/picamctl.sh
akseidel: In combination with auto discover the application Webcamoid is run for Linux situations. This has been tested on a Raspberry Pi 4 running a Raspbian system.

I could be wrong but xdg-open wants an image file for an argument, so I did not find a way to use it. In the script situation an application needs to be run, For the Linux situation I found running a detached  screen running Webcamoid as a solution. The difference between the Apple and Linux situations is this detached screen existence that gets blown away the next time the script is run. Webcamoid closes and reopens if it was already running whereas an already running PhotoBooth just stays.

picamctl/picamctl.sh
akseidel: Based on my tests with a Raspberry Pi 4, an even longer sleep is required, particularly this first one immediately after starting the screen connection.

kbingham:
I'm curious about this, is it to be able to configure the camera when it's running? Wouldn't that be handled by setting controls on the host which are interpreted by uvc-gadget and then set accordingly?

Or is this more specific?

The camera should be configurable without using the serial console.

dgsiegel:
> The camera should be configurable without using the serial console.

Yes, you're right, see https://github.com/showmewebcam/showmewebcam/pull/93

However this is not fully functional right now, as it needs kernel/uvc patches and probably further changes in uvc-gadget.

dgsiegel:
#109 might be interesting to you, as it enables auto login through the serial console

akseidel:
Yes the auto login looks promising. I was planning to eliminate the greeting line myself as the very first change to make if I ever managed to venture into building it.

picamctl/picamctl.sh
akseidel: Now I know what you are talking about. I would never see that on my Pi 4 where I run as root. The script now checks for root user and advises running the script with sudo. It also checks for a screen command.

PR #101, opened 2021-02-10
raspberrypi4: initial support

calebccff:
This PR is loosely based on !49

Some unfortunately major changes seem to be needed to make this work.

The kernel has been updated to the latest 5.10.y tag on raspberrypi/linux (5.10.11).

The pi4 uses a very different kernel config to the pi0, the hardware is
also very different so using a base config doesn't seem like the
sensible approach here.

A future improvement would be to have a common config fragment
containing only the config options needed for uvc gadget that aren't
included in the defconfig, as well as perhaps a board-specific fragment,
the build script could then take the stock defconfig, merge in the board
specific overlay and the showmewebcam specific overlay and use that as
the kernel config.

The modifications to the build script change up the format to use a per-board defconfig, with and optional overlay, the idea being that the defconfig is pulled straight from raspberrypi/linux.

The file `board/linux-raspberrypi4.config` is straight from !49.

I am not currently able to test a pi zero image.


dgsiegel:
First of all, thanks for working on this!

> The kernel has been updated to the latest 5.10.y tag on raspberrypi/linux (5.10.11).

Generally I don't have anything against upgrading the kernel version. Is there a reason why you need that specific version? 

Also, I'd prefer to use a commit id instead of a tag, as these don't seem to follow a specific release cycle.

> The modifications to the build script change up the format to use a per-board defconfig, with and optional overlay, the idea being that the defconfig is pulled straight from raspberrypi/linux.

That won't work. We've spent lots of time optimizing the kernel config to speed up boot time, hence falling back to a defconfig is not the way to go. I'd be happy though to reduce `linux-base.config` to a minimum which is usable by all platforms. Overlays (`board/linux-${BOARDNAME}.config`) are already possible with the current setup, we don't need a third layer here.

configs/config
```diff
@@ -1,7 +1,6 @@
 BR2_arm=y
-BR2_arm1176jzf_s=y
 BR2_TOOLCHAIN_BUILDROOT_GLIBC=y
-BR2_KERNEL_HEADERS_5_4=y
+BR2_PACKAGE_HOST_LINUX_HEADERS_CUSTOM_5_9=y
```
dgsiegel, line 3: Why not `BR2_KERNEL_HEADERS_5_9`?

board/genimage-raspberrypi4.cfg
```diff
@@ -0,0 +1,32 @@
...
+      "rpi-firmware/config.txt",
+      "rpi-firmware/fixup.dat",
+      "rpi-firmware/start.elf",
+      "rpi-firmware/overlays",
+      "camera.txt",
+      "enable-serial-debug",
+      "zImage"
+    }
+  }
+  size = 16M
```
dgsiegel, line 15: Is 16M enough for the Pi4 as well?

board/linux-raspberrypi0.config
```diff
@@ -15,9 +15,12 @@ CONFIG_IKCONFIG=m
 CONFIG_IKCONFIG_PROC=y
 CONFIG_MEMCG=y
 CONFIG_BLK_CGROUP=y
+CONFIG_CFS_BANDWIDTH=y
```
dgsiegel, line 18: See my previous comment, we don't need all these additional settings and modules

configs/config
```diff
@@ -14,7 +13,7 @@ BR2_ROOTFS_POST_IMAGE_SCRIPT="$(BR2_EXTERNAL_PICAM_PATH)/board/post-image.sh"
 BR2_ROOTFS_POST_SCRIPT_ARGS="--configure-picam"
 BR2_LINUX_KERNEL=y
 BR2_LINUX_KERNEL_CUSTOM_TARBALL=y
-BR2_LINUX_KERNEL_CUSTOM_TARBALL_LOCATION="$(call github,raspberrypi,linux,c5e512a329ac7a0bfdd5f53477f4300723618db5)/linux-c5e512a329ac7a0bfdd5f53477f4300723618db5.tar.gz"
+BR2_LINUX_KERNEL_CUSTOM_TARBALL_LOCATION="$(call github,raspberrypi,linux,raspberrypi-kernel_1.20210201-1)/linux-raspberrypi-kernel_1.20210201-1.tar.gz"
```
dgsiegel, line 16: commit id would be better than a tag

build-showmewebcam.sh
```diff
@@ -26,11 +30,13 @@ esac
 CONFIG_="BR2" KCONFIG_CONFIG="configs/${BOARDNAME}_defconfig" "$BUILDROOT_DIR/support/kconfig/merge_config.sh" -m -r configs/config "configs/$BOARDNAME"
 sed "1i ### DO NOT EDIT, this file was automatically generated\n" -i "configs/${BOARDNAME}_defconfig"
 
-# Merge kernel configurations
-if [ -f "board/linux-${BOARDNAME}.config" ]; then
```
dgsiegel, line 30: Overlays would be possible with this. The better approach would be to make linux-base.config usable for the Pi4 as well.

calebccff:
> First of all, thanks for working on this!
> 
> > The kernel has been updated to the latest 5.10.y tag on raspberrypi/linux (5.10.11).
> 
> Generally I don't have anything against upgrading the kernel version. Is there a reason why you need that specific version?
> 
> Also, I'd prefer to use a commit id instead of a tag, as these don't seem to follow a specific release cycle.
> 
> > The modifications to the build script change up the format to use a per-board defconfig, with and optional overlay, the idea being that the defconfig is pulled straight from raspberrypi/linux.
> 
> That won't work. We've spent lots of time optimizing the kernel config to speed up boot time, hence falling back to a defconfig is not the way to go. I'd be happy though to reduce `linux-base.config` to a minimum which is usable by all platforms. Overlays (`board/linux-${BOARDNAME}.config`) are already possible with the current setup, we don't need a third layer here.

I see, so currently you've minimised the base config, however that config does not work on the pi4, and I'm not sure if it's possible to have a single config that will work for the pi zero and 4.

The reason I proposed a 3rd layer was so that the minimisation you do can be applied to both devices, you can use the original defconfig, showmewebcam optimisations and if necessary a board specific overlay config. That 3rd one probably wouldn't be needed. That way most of the optimisations just happen, and new boards don't have to manually trim down their kernel config.

Basically, I think it would make sense to make the config optimisations generic, the alternative would be me manually optimisation the pi4 defconfig too, and any future devices having to do the same as well, my Pi 4 boots comfortably fast in the current configuration so I'd be happy disabling modules and leaving it at that.

configs/config
calebccff: This is actually my first time touching buildroot, `KERNEL_HEADERS` is headers used for compiling for device, and `PACKAGE_HOST_LINUX_HEADERS` are the headers used for compiling host tools?

I'll change to your recommendation, that seems to actually make sense :sweat_smile:

board/genimage-raspberrypi4.cfg
calebccff: I haven't had any issues with it so I would presume so, this was taken from !49 and I think is just the same as the Pi zero. Perhaps worth increasing just to be safe?

calebccff:
> Generally I don't have anything against upgrading the kernel version. Is there a reason why you need that specific version?

I think I read somewhere that 5.4 has broken gadget mode on the Pi 4, 5.10 seemed like a sensible pick.

build-showmewebcam.sh
calebccff: Like I mentioned in my comment, I don't think it's feasible to use the same base config for the pi zero and pi 4. They use different generations of SoC, the pi zero is Armv6 and the Pi 4 is v8 (although here it's only running as armv7l). Making an overlay for the base config currently in master would be disabling a lot of SoC specific options and enabling others which would be a lot more likely to hit issues down the line, as well as a lot more confusing to maintain.

Flipping that idea on it's head, using the defconfig for the board, and a common overlay to disable useless modules and builtins means that the same optimisations don't need to be done by hand for every board.

kbingham:
Hi Caleb, I've tried this image, and found that the Serial gadget comes up, but not the UVC gadget.
Looking at the logs with systemctl, I can see:
systemctl status usb-gadget-config

> ● usb-gadget-config.service - Configure Raspberry Pi as a USB gadget
>      Loaded: loaded (/usr/lib/systemd/system/usb-gadget-config.service; enabled; vendor preset: enabled)
>      Active: inactive (dead) since Sat 2020-09-12 12:16:34 UTC; 9min ago
>     Process: 142 ExecStart=/opt/uvc-webcam/multi-gadget.sh (code=exited, status=0/SUCCESS)
>    Main PID: 142 (code=exited, status=0/SUCCESS)
> 
> Sep 12 12:16:33 piwebcam systemd[1]: Starting Configure Raspberry Pi as a USB gadget...
> Sep 12 12:16:33 piwebcam usb-gadget-config[142]: I did not detect a camera connected to the Pi. Please check your hardware.
> Sep 12 12:16:33 piwebcam usb-gadget-config[142]: Configuring USB gadget serial interface
> Sep 12 12:16:34 piwebcam systemd[1]: usb-gadget-config.service: Succeeded.
> Sep 12 12:16:34 piwebcam systemd[1]: Finished Configure Raspberry Pi as a USB gadget.

However, looking on the device itself, I can clearly see /dev/video0 is there.
I wonder if this is a race between the usb-gadget-config.service running before the camera has fully been loaded.

Did you experience this at all?

I can see there is already a `udevadm settle` at the end of /opt/uvc-webcam/multi-gadget.sh, but I suspect it should probably also wait for udev to settle *before* looking to see if the video camera is already there.

kbingham:
Aha, so I see you have a commit today that references this issue in fact, with a different solution.
I'll test that image now.

calebccff:
> Aha, so I see you have a commit today that references this issue in fact, with a different solution.
> I'll test that image now.
@kbingham 
Hi, there seems to be some kernel config related issue here, in `linux-raspberrypi4-overlay.config` with the values set to =n they override the previous value, but if they're commented out they don't. I tested them with `# CONFIG_XYZ is not set` but changed it to be `CONFIG_XYZ=n` before pushing, my guess is one of them being disable breaks the camera on the pi side.

kbingham:
I tested sdcard-raspberrypi4-9ddee85.img which loaded the camera just fine. I think the camera issue I saw was just about the boot speed of the rpi4. It's running the boot scripts before the webcam has fully loaded. The delay fixed it, and I suspect a udevadm settle before checking the camera is there would also work.

package/piwebcam/multi-gadget.sh
```diff
@@ -86,14 +86,27 @@ config_usb_webcam () {
 }
 
 # Check if camera is installed correctly
-if [ ! -e /dev/video0 ] ; then
-  echo "I did not detect a camera connected to the Pi. Please check your hardware."
-  CONFIGURE_USB_WEBCAM=false
+if [ -e /dev/video0 ] ; then
+  CONFIGURE_USB_WEBCAM=true
+else
+  COUNT=0
```
kbingham, line 92: I suspect all of this delay loop could be simplified by calling 'udevadm settle' before 
" if [ ! -e /dev/video0 ] ; then"

package/piwebcam/multi-gadget.sh
```diff
@@ -86,14 +86,27 @@ config_usb_webcam () {
...
+      COUNT=$((COUNT+1))
+  done
 fi
 
 if [ "$CONFIGURE_USB_WEBCAM" = true ] ; then
   echo "Configuring USB gadget webcam interface"
   config_usb_webcam
+else
+  echo "No camera detected after 3 seconds. Enabled serial for debugging."
+  CONFIGURE_USB_SERIAL=true
```
kbingham, line 109: That's a good idea though - if the camera doesn't load, it should force the serial to be enabled I think.

package/piwebcam/multi-gadget.sh
calebccff: That would 100% be a better solution.

Although we're running systemd so perhaps we can go even further... In this case just waiting for https://www.freedesktop.org/software/systemd/man/systemd-udev-settle.service.html to finish should be enough as we assume the camera is attached before the pi is booted, or not at all.

kbingham: ah yes, if there's a systemd way to do this then that's even better indeed.

build-showmewebcam.sh
dgsiegel: As I said before, defconfig is not the way to go, as we'll loose all optimizations. Flipping this on its head won't work either, as we'll always chasing new and changed options when upgrading the kernel. Hence the better way, still, is to have a base linux config and add all further changes to their variants. If that means moving SoC specific options to their corresponding files, I won't have any problems with that.

package/piwebcam/multi-gadget.sh
dgsiegel: systemd can also wait for a device to appear, both ways are fine.

package/piwebcam/multi-gadget.sh
dgsiegel: There's a good chance that serial won't work, as uvc-gadget fails. Didn't test this for this specific case though.

skarard:
I ran into the same problems as @calebccff.

I ended up going for a linux.config file that, when merged into the defconfig of a board, would add and strip out the modules for boot optimisation. It works for pi4 and no reason why not for the zero but untested. 

I tried to merge the diff of the the resulting pi4 and zero linux.configs but I guess the zero had a package that overwrites some default package that it required by the pi4 but the pi4 defconfig doesn't specify. I couldn't find what that was. 

Have a peek at https://github.com/skarard/pi4shadercam/tree/pi4-support if that's helpful.

package/piwebcam/multi-gadget.sh
skarard: It could be useful at least for debugging camera issues. If you don't have uvc then you have bigger issues. Say your camera cable came loose or your camera didn't register in time this can help debug were the problem is.

kbingham: > There's a good chance that serial won't work, as uvc-gadget fails. Didn't test this for this specific case though.

No, it's fine. If the configfs isn't set up (because config_usb_webcam isn't called) then there will be nothing blocking in the composite gadget, and it won't matter if uvc-gadget isn't running.

calebccff: @dgsiegel I just tested with the latest version of commits here (99bf203) and you were right, with no camera attached the device doesn't register properly.

Even when it does appear as a serial gadget it's been near unusable for me, constantly freezing up often before I can even log in. Is it worth switching to something like rndis network gadget instead? Although that does come with a little more overhead.

Then again, it may well just be some missing config option.

package/piwebcam/multi-gadget.sh
```diff
@@ -101,5 +103,4 @@ if [ "$CONFIGURE_USB_SERIAL" = true ] ; then
   config_usb_serial
 fi
 
-udevadm settle -t 5 || :
```
kbingham, line 104: Can this be removed?

Creating the usb-composite gagdets will create new devices in the kernel, and that will create udev events. It might be that the udev settle here is used to make sure we wait for those to exist before we move on ...

I don't think I've ever needed to do a udevadm settle here after creating the devices, but I probably normally ran things from the command line, so there could be races.

calebccff: I didn't consider that, there's no real reason /not/ to keep it there so probably worth adding back.

package/piwebcam/multi-gadget.sh
kbingham: That's interesting, the serial gadget worked well for me when I used it this week.
But the freezing/lockups sound similar to what I was experiencing while testing gadget (outside of show-me-webcam)

calebccff: I'll have to do some testing... rndis has been a much more reliable experience than USB_G_SERIAL for me in the past. I've just made the script enable serial if the camera isn't detected again, so I'll mark this as resolved.

board/linux-raspberrypi0.config
calebccff: Can we look at trying to get this merged? It seems like we've hit a bit of a brick wall with the kernel configs, I'm happy to have a crack at reducing the pi4 config as there is a lot of room for improvement there. It will take some help from you guys to find a good solution here as I'm not too familiar with the Raspberry Pi kernel config.

dgsiegel: As I said many times already, I don't think the overlay approach is the right one. The changes needed would be quite simple, actually:

- Diff the current Pi0/w and Pi4 configs and look for the same lines. These will go into `linux-base.config`
- Put all other configs into their respective `board/linux-${BOARDNAME}.config` config

Ideally, the resulting config for the Pi0/w should stay the same, even with this PR merged.

build-showmewebcam.sh
kbingham: Is there anything that details what optimisations have been made already to the rpi0?

Would you envisage this being a common base against other board targets too, for example when using a Rockchip RK3399? or a Qualcomm target?

Going to AARCH64 may also have quite an impact on the kernel configurations, and of course all of this has to be maintainable as the kernel version itself is bumped to newer releases?

dgsiegel: > Is there anything that details what optimisations have been made already to the rpi0?

Kind of here: https://github.com/showmewebcam/showmewebcam/commit/b680695687099045befdce3174a4c6db4f7166d2 and https://github.com/showmewebcam/showmewebcam/commit/a7e933e20ff4e52a965b77616ab64c1b36448e27

> Would you envisage this being a common base against other board targets too, for example when using a Rockchip RK3399? or a Qualcomm target?

I don't see why not

> Going to AARCH64 may also have quite an impact on the kernel configurations, and of course all of this has to be maintainable as the kernel version itself is bumped to newer releases?

Exactly. Hence the approach of using config stubs and merging them with the default buildroot kernel settings.

calebccff:
Ok, resolved the camera device not being detected until AFTER gadget setup for reals this time!
Making the following built in is the solution:
```
CONFIG_MEDIA_SUPPORT=y
CONFIG_VIDEO_BCM2835=y
CONFIG_I2C_BCM2835=y
CONFIG_VIDEO_CODEC_BCM2835=y
CONFIG_VIDEO_BCM2835_UNICAM=y
```

@dgsiegel I can only imagine this improving the boot time of the Pi Zero too, I'll keep these in the Pi4 config fragment and leave it up to you guys if you think it's worth including in the Pi Zero as well (if these drivers are used there too).

kbingham:
Hi Caleb,

CONFIG_MEDIA_SUPPORT=y
CONFIG_VIDEO_BCM2835=y
CONFIG_I2C_BCM2835=y
CONFIG_VIDEO_CODEC_BCM2835=y
CONFIG_VIDEO_BCM2835_UNICAM=y

Could you test with the BCM2835_UNICAM disabled please? That isn't used, and shouldn't be needed in this platform currently.
(That will be used by the libcamera stack though, but isn't needed here currently).

In fact, looking at that list you might be able to chop it down to just 
CONFIG_VIDEO_BCM2835=y

That's the driver used to handle the camera in this case.

It might need MEDIA_SUPPORT, but I think that's only a dependency on the CODEC_BCM2835 from when I added media controller support to those drivers.

Even the I2c support might not be needed here, as the I2C for the camera is handled through the VC4/GPU, not the ARM

calebccff:
I'll split the gadget fixes into a separate PR.

build-showmewebcam.sh
calebccff: OK, latest version should follow this idea pretty well. Resolving this.

htruong:
Awesome! I will let @dgsiegel approve it, since he is more involved with this PR, but it looks good to me.

dgsiegel (approved):

dgsiegel:
Merged. @calebccff thanks so much!

PR #113, opened 2021-03-07
Multi gadget fixes

calebccff:
Ensure udev has settled before we setup the gadget, on the Pi 4 at least this is nice to have, although probably not needed (when camera support is built as a module I've still had it show up late).

Also make sure to enable serial so that logs can be retrieved if we fail to find a camera.


htruong (approved):

htruong:
Thanks for the fix, it looks good to me.

dgsiegel:
Tested it on the Pi0 as well. Thanks so much for contributing!

Issue #116, opened 2021-03-22
UVC query errors


dgsiegel:
When booting up a Pi0 (probably happens with Pi0w and Pi4 as well), I get this in `dmesg`:

```
[  +0,007594] uvcvideo: Found UVC 1.00 device Piwebcam (1d6b:0104)
[  +0,004627] uvcvideo: Failed to query (GET_INFO) UVC control 6 on unit 2: -32 (exp. 1).
[  +0,007948] uvcvideo: Failed to query (GET_INFO) UVC control 9 on unit 2: -32 (exp. 1).
[  +0,002236] uvcvideo: Failed to query (GET_INFO) UVC control 10 on unit 2: -32 (exp. 1).
[  +0,003767] uvcvideo: Failed to query (GET_INFO) UVC control 4 on unit 2: -32 (exp. 1).
[  +0,009293] input: Piwebcam: UVC Camera as /devices/pci0000:00/0000:00:14.0/usb1/1-4/1-4.1/1-4.1.2/1-4.1.2:2.0/input
/input57
```

Similarly, when using `v4l2-ctl -d /dev/video2 -l` you can trigger the same messages in dmesg:

```
[  +3,491996] uvcvideo: Failed to query (GET_DEF) UVC control 6 on unit 2: -32 (exp. 2).
[  +0,007836] uvcvideo: Failed to query (GET_DEF) UVC control 6 on unit 2: -32 (exp. 2).
```

These failed queries look like they come from this commit: 8ba054b25ef8a5c2993473088795deac4588c5e4

@tuyenld could you have a look into this?

tuyenld:
I'm a little busy right now, I'll take a look at weekend.

tuyenld:
I can't reproduce this message in `dmesg`. I tested with Pi Zero + v1.3 Camera on Debian.

Anyway, I think it is safe to ignore these error messages. My commit enables video controls for many type of camera as much as possible.
However, not every camera hardware supports these controls. 

For example, [Pi camera v1.3 OV5647](https://cdn.sparkfun.com/datasheets/Dev/RaspberryPi/ov5647_full.pdf) didn't support `White Balance Temperature`, but Rasp does report that it support this control , then, you got this error:
```
uvcvideo: Failed to query (GET_INFO) UVC control 6 on unit 2: -32 (exp. 1).
```

dgsiegel:
It definitely happens with the HQ cam. Is there a way we could only enable the right controls for the right camera module?

[...]

Issue #50, opened 2020-11-27
brightness setting in camera.txt is being ignored


zlorfi:
After writing `/boot/camera.txt` with
```
#Tweak the auto exposure bias
auto_exposure_bias=15
#Tweak the brightness
brightness=55
```
the `auto_expose_bias` value is being honoured, but not the `brightness` value. 
```
# /usr/bin/v4l2-ctl -L

User Controls

             brightness 0x00980900 (int)    : min=0 max=100 step=1 default=50 value=50 flags=slider
[...]          
             auto_exposure_bias 0x009a0913 (intmenu): min=0 max=24 default=12 value=15
                                0: -4000 (0xfffffffffffff060)
                                1: -3667 (0xfffffffffffff1ad)
                                2: -3333 (0xfffffffffffff2fb)
                                3: -3000 (0xfffffffffffff448)
                                4: -2667 (0xfffffffffffff595)
                                5: -2333 (0xfffffffffffff6e3)
                                6: -2000 (0xfffffffffffff830)
                                7: -1667 (0xfffffffffffff97d)
                                8: -1333 (0xfffffffffffffacb)
                                9: -1000 (0xfffffffffffffc18)
                                10: -667 (0xfffffffffffffd65)
                                11: -333 (0xfffffffffffffeb3)
                                12: 0 (0x0)
                                13: 333 (0x14d)
                                14: 667 (0x29b)
                                15: 1000 (0x3e8)
                                16: 1333 (0x535)
                                17: 1667 (0x683)
                                18: 2000 (0x7d0)
                                19: 2333 (0x91d)
                                20: 2667 (0xa6b)
                                21: 3000 (0xbb8)
                                22: 3333 (0xd05)
                                23: 3667 (0xe53)
                                24: 4000 (0xfa0)
[...]
```

Any idea how to fix this problem?
In order to have the brightness at `55` after a reboot, I have to set it manually via `/usr/bin/v4l2-ctl -c brightness=55` which seems to work until the next reboot.

dgsiegel:
I can't seem to reproduce this. 

# v4l2-ctl -l
[...]
             brightness 0x00980900 (int)    : min=0 max=100 step=1 default=50 value=55 flags=slider
[...]
             auto_exposure_bias 0x009a0913 (intmenu): min=0 max=24 default=12 value=15

Could you paste your full `camera.txt` file, as well as the output of `journalctl`?

zlorfi:
```bash
[...]

PR #118, opened 2021-03-30
Set processing controls from cmdline

wherget:
This PR moves the values for usb descriptor `bmControls` bitmap to a commandline parameter, so we can experiment what host OSes expect to see there and what actually works with the Pi cameras, without having to recompile a kernel all the time.

The Bitmap can be changed by editing `cmdline.txt` on the SD to say `usb_f_uvc.bm_controls=<byte1>,<byte2>`.

Submitting as Draft to get it out there, might not yet work on master since it might need support from [uvc-gadget](/peterbay/uvc-gadget), but should help against the issues seen in #116

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,6 +153,28 @@ You can edit `camera.txt` on-target by remounting `/boot` read-write:
 mount -o remount,rw /boot
 ```
 
+### Overriding the controls available to the host computer
+
+Since version 1.80 we try to expose as many controls as possible via the UVC
+standard, which are then translated back to the controls that are available
+on your specific camera module. Some host operating systems may be confused
+by controls that are advertised via USB, but don't turn out to work, due to
+not being available for your specific camera module.
+
+You can therefore customize the controls advertised to the computer via the
+USB device descriptor. *Warning*: This is an advanced user feature only.
+You should probably consult the [UVC controls bitmap documentation] on GitHub.
+
+You can add the parameters to `cmdline.txt` on the boot volume as follows:
+
+```
+usb_f_uvc.camera_terminal_controls=0,0,0 usb_f_uvc.processing_unit_controls=0,0
+```
+
+The previous example sets all controls to disabled, and should thus be safe.
+The parameters directly correspond to the `bmControls` bitfields in the descriptor.
+Please, again, read the documentation linked above.
+
 ## Development & building
 
 Clone or download this repository. Then inside it:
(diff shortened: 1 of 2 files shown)


dgsiegel:
Looks quite nice already! Just wondering whether support for this should be in uvc-gadget or not?

wherget:
I've updated the PR to include a parameter for `camera_controls` (the other got renamed to `proc_controls`).

I think this might be OK to merge now functionality-wise, I'd appreciate it if someone beside me could test it though.
Where should documentation for this feature go? It's not particularly hard to use, so could go into README, but then again computing the right bitfield values isn't exactly end-user friendly. 

In the long run, I think we'd be better off to add write capability to the configfs interface anyway.

wherget:
Linking to relevant issues, looks like this might help people in #50 (by disabling controls entirely) and #122 (helping to debug what effects the controls bitmap has).

dgsiegel:
> I've updated the PR to include a parameter for camera_controls (the other got renamed to proc_controls).

Just wondering if we should keep those as close to the original names? E.g. `uvc_camera_terminal` and `uvc_processing`?

`usb_f_uvc.proc_controls=165,0 usb_f_uvc.camera_controls=0,0,0` seems to disable all host settings:
`
$ v4l2-ctl -d /dev/video2 -l

`
> In the long run, I think we'd be better off to add write capability to the configfs interface anyway.

Agree, maintaining a patch isn't the best way forward. Ideally we could enable everything with this patch and then control the individual settings via uvc-gadget. What do you think?

So, is this ready to be merged?

dgsiegel (dismissed):

wherget:
> Just wondering if we should keep those as close to the original names?

I agree on keeping the names close to what the parameters affect. The only previously abailable module option is called `trace`, and I would tend to side with the original author's choice of dropping the `uvc_` prefix, as `usb_f_uvc.uvc_blah` reads somewhat redundant. That would give us (quickly glances at the USB standard) `camera_terminal_controls` and `processing_unit_controls`. That seems a bit on the long side for module parameters, but I wouldn't mind. What do you think?

> Ideally we could enable everything with this patch and then control the individual settings via uvc-gadget.

If we were to have configfs write to the `bmControls` attribute(s) that would make the entire patch redundant. We could just recompute the bitmap from the controls available on the upstream video device and write it to configfs. Not sure about `uvc-gadget` being the right place, IIRC that presumes the gadget video device to be set up and bound already before launch, and `bmControls` would (presumably) have to be set before the host enumerates the device. But yeah, something of sorts.

"Enable everything and let uvc-gadget handle only the controls available" (without updating bmControls from configfs) is the state before this PR, had me confused there for a second.

> So, is this ready to be merged?

I'd like to:
- [ ]  add a notice to the README about the existence of this feature, with at least a link to further documentation in the Wiki, where we can more easily collect "correct" values for different setups.
- [ ] rename the parameters (again, heh) to a sensible scheme as per above discussion.

After that it most definitely can be merged.

dgsiegel (approved):

dgsiegel:
Awesome, let's do it! Thanks so much for all your effort and insights!

Issue #90, opened 2021-01-12
Resolution with Teams on Windows 10 is too low


westfeld:
Hello everyone,

I am experiencing an issue when using the webcam on a Windows 10 PC with Teams. The resolution is not set to 1080p when using the camera. I checked this using the serial interface and camera-ctl.

**After startup:**
Resolution: 1024x768
Format: JPEG

**When using Teams (preview camera in Teams' settings and in an actual video meeting)**
Resolution: 640x360
Format: MJPG

Right now my workaround is to use OBS and the virtual camera to loop the video through it and indeed I get a nice 1080p HD output from the camera. It would be easier to just do the right thing and not having to fiddle around with OBS.

rdcuk:
I have the same experience.

Am unable to use the Windows Camera app trick mentioned elsewhere. Really need the ability to override the resolution which teams wants. Could it be added to config somewhere?

I know we're effectively making up for the lack of settings in Teams here, but I guess we're all stuck with it :-(

dgsiegel:
We currently set up all resolutions in `package/piwebcam/multi-gadget.sh`. It would be easy to outsource those into a config file in /boot and make them configurable there. Feel free to send a PR.

rdcuk:
Sounds good. I wouldn't really know where to start on this myself (unless you want it re-impementing in COBOL), but if anyone has the time/inclination then I would be happy to test.

westfeld:
I have now rebuilt a new variant of Showmewebcam and just left out the following resolutions in `package/piwebcam/multi-gadget.sh`

    -  config_frame mjpeg m  640  360
    -  config_frame mjpeg m  640  480
    -  config_frame mjpeg m  800  600
    -  config_frame mjpeg m 1024  768


At first sight it looks better, but I will test that. Once it has been confirmed that it works better, we can think about making it configurable more easily.

westfeld:
After more testing with Teams on Windows 10, it works flawlessly in 720p resolution.

rdcuk:
Is there any way to change this on the standard SD card? i.e. without building a specific image on another system

dgsiegel:
No, but we'd be happy to accept a patch making this configurable, e.g. with a file in /boot.

PR #128, opened 2021-04-17
Add configurable resolutions

westfeld:
I need some help where to put the file `resolutions.txt` in the build tree to be included in the `/boot` directory

adresses Issue #90

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,6 +171,51 @@ The previous example sets all controls to disabled, and should thus be safe.
 The parameters directly correspond to the `bmControls` bitfields in the descriptor.
 Please, again, read the documentation linked above.
 
+### Configure available camera resolutions and streaming format
+
+The camera advertizes to the connected host PC the available resolutions and
+streaming formats of the video signal. Usually the best suitable resolution is
+chosen. The available resolutions can be reduced to enforce a specific setting.
+
+By default the camera uses the *Motion JPEG* codec to stream the video signal.
+For the host PC it offers several video resolutions up to 1080p HD video to
+choose from.
+
+These settings can be overwritten by copying the file `video_formats.txt` into
+the `/boot` folder and edit it.
+
+```bash
+cp /etc/video_formats.txt /boot
+```
+
+In this file one line corresponds to one video resolution and streaming format.
+Until now, only mjpeg and uncompressed video are supported. If you want to make
+the lower resolutions up to 720p HD unavailable, you have to comment out the
+first 4 resultions by prepending a `#` to that line.
+
+Please note that you have to reboot your piwebcam (power cyle it) for changes
+to take effect.
+
+Content of `video_formats.txt`
+```
+# Default Show-me webcam supported resolutions
+#
+# format: video format(mjpeg|uncompressed) x-dimension y-dimension
+#
+mjpeg 640 360
+mjpeg 640 480
+mjpeg 800 600
+mjpeg 1024 768
+mjpeg 1280 720
+mjpeg 1280 960
+mjpeg 1440 1080
+mjpeg 1536 864
+mjpeg 1600 900
+mjpeg 1600 1200
+mjpeg 1920 1080
+```
+ 
+
 ## Development & building
 
 Clone or download this repository. Then inside it:
--- a/package/piwebcam/multi-gadget.sh
+++ b/package/piwebcam/multi-gadget.sh
@@ -10,6 +10,11 @@ if [ -f "/boot/enable-serial-debug" ] ; then
   CONFIGURE_USB_SERIAL=true
 fi
 
+VIDEO_FORMATS_FILE=/etc/video_formats.txt
+
+# location of video_formats.txt file if overwritten by the user
+VIDEO_FORMATS_USER_FILE=/boot/video_formats.txt
+
 CONFIG=/sys/kernel/config/usb_gadget/piwebcam
 mkdir -p "$CONFIG"
 cd "$CONFIG" || exit 1
@@ -64,27 +69,29 @@ EOF
 config_usb_webcam () {
   mkdir -p functions/uvc.usb0/control/header/h
 
-  config_frame mjpeg m  640  360
-  config_frame mjpeg m  640  480
-  config_frame mjpeg m  800  600
-  config_frame mjpeg m 1024  768
-  config_frame mjpeg m 1280  720
-  config_frame mjpeg m 1280  960
-  config_frame mjpeg m 1440 1080
-  config_frame mjpeg m 1536  864
-  config_frame mjpeg m 1600  900
-  config_frame mjpeg m 1600 1200
-  config_frame mjpeg m 1920 1080
+  if [ -r $VIDEO_FORMATS_USER_FILE ] ; then
+      FORMATS_FILE=$VIDEO_FORMATS_USER_FILE
+  else
+      FORMATS_FILE=$VIDEO_FORMATS_FILE
+  fi
 
   mkdir -p functions/uvc.usb0/streaming/header/h
-  ln -s functions/uvc.usb0/streaming/mjpeg/m  functions/uvc.usb0/streaming/header/h
   ln -s functions/uvc.usb0/streaming/header/h functions/uvc.usb0/streaming/class/fs
   ln -s functions/uvc.usb0/streaming/header/h functions/uvc.usb0/streaming/class/hs
   ln -s functions/uvc.usb0/control/header/h   functions/uvc.usb0/control/class/fs
 
   ln -s functions/uvc.usb0 configs/c.2/uvc.usb0
-}
 
+   grep -E "^(mjpeg|uncompressed)\s+[[:digit:]]+\s+[[:digit:]]+" $FORMATS_FILE | sed -E "s/\s+/ /g" | while read -r line
+  do
+    VIDEO_FORMAT=$(echo "$line" | cut -d ' ' -f1)
+    HDR_DESC=$(echo "$VIDEO_FORMAT" | cut -c 1)
+    X=$(echo "$line" | cut -d ' ' -f2)
+    Y=$(echo "$line" | cut -d ' ' -f3)
+    config_frame "$VIDEO_FORMAT" "$HDR_DESC" "$X" "$Y"
+    ln -sf "functions/uvc.usb0/streaming/$VIDEO_FORMAT/$HDR_DESC"  functions/uvc.usb0/streaming/header/h
+  done
+} 
 # Check if camera is installed correctly
 if [ ! -e /dev/video0 ] ; then
   echo "I did not detect a camera connected to the Pi. Please check your hardware."
--- a/package/piwebcam/piwebcam.mk
+++ b/package/piwebcam/piwebcam.mk
@@ -21,6 +21,7 @@ define PIWEBCAM_INSTALL_TARGET_CMDS
 	$(INSTALL) -D -m 0755 $(@D)/uvc-gadget $(TARGET_DIR)$(PIWEBCAM_DEST_DIR)
 	$(INSTALL) -D -m 0755 $(PIWEBCAM_PKGDIR)/multi-gadget.sh $(TARGET_DIR)$(PIWEBCAM_DEST_DIR)
 	$(INSTALL) -D -m 0755 $(PIWEBCAM_PKGDIR)/start-webcam.sh $(TARGET_DIR)$(PIWEBCAM_DEST_DIR)
+	$(INSTALL) -D -m 0644 $(PIWEBCAM_PKGDIR)/video_formats.txt $(TARGET_DIR)/etc/video_formats.txt
 endef
 
 define PIWEBCAM_INSTALL_INIT_SYSTEMD
--- a/package/piwebcam/video_formats.txt
+++ b/package/piwebcam/video_formats.txt
@@ -1,0 +1,16 @@
+# Default Show-me webcam supported resolutions
+#
+# format: video format(mjpeg|uncompressed) x-dimension y-dimension
+#
+mjpeg 640 360
+mjpeg 640 480
+mjpeg 800 600
+mjpeg 1024 768
+mjpeg 1280 720
+mjpeg 1280 960
+mjpeg 1440 1080
+mjpeg 1536 864
+mjpeg 1600 900
+mjpeg 1600 1200
+mjpeg 1920 1080
+ 


htruong:
Straightforward and very clean. LGTM. @dgsiegel ?

westfeld:
Thanks for the comment, still I need help how to place the `resolutions.txt` into `/boot`.

package/piwebcam/resolutions.txt
```diff
@@ -0,0 +1,16 @@
+# Default Show-me webcam supported resolutions
+#
+# uncomment resolutions if not wanted
```
mseri, line 3: Shouldn’t this be “comment”?

westfeld: Yes, you are right - commited a fix in 19dc199

westfeld:
Fixes #90

dgsiegel:
Looks very nice, Thomas! Thanks for your work. I'll add a few comments to the patch directly, but would like to merge this as soon as possible and create a new release.

board/post-image.sh
```diff
@@ -90,6 +90,11 @@ __EOF__
 			cat "$BR2_EXTERNAL_PICAM_PATH/camera.txt" >> "${BINARIES_DIR}/camera.txt"
 		fi
 
+		# Add default resolutions.txt if it exists
+		if [ -f "$BR2_EXTERNAL_PICAM_PATH/package/piwebcam/resolutions.txt" ]; then
```
dgsiegel, line 94: I don't think this should be an optional file, like `camera.txt` is. We could certainly move this file to `/etc`.

In that case I *think* you could move that line to `package/piwebcam/piwebcam.mk`?

package/piwebcam/multi-gadget.sh
```diff
@@ -64,17 +66,26 @@ EOF
...
+    config_frame mjpeg m  640  480
+    config_frame mjpeg m  800  600
+    config_frame mjpeg m 1024  768
+    config_frame mjpeg m 1280  720
+    config_frame mjpeg m 1280  960
+    config_frame mjpeg m 1440 1080
+    config_frame mjpeg m 1536  864
+    config_frame mjpeg m 1600  900
+    config_frame mjpeg m 1600 1200
+    config_frame mjpeg m 1920 1080
```
dgsiegel, line 87: See above, this could be a global configuration setting

package/piwebcam/multi-gadget.sh
```diff
@@ -64,17 +66,26 @@ EOF
...
-  config_frame mjpeg m 1024  768
-  config_frame mjpeg m 1280  720
-  config_frame mjpeg m 1280  960
-  config_frame mjpeg m 1440 1080
-  config_frame mjpeg m 1536  864
-  config_frame mjpeg m 1600  900
-  config_frame mjpeg m 1600 1200
-  config_frame mjpeg m 1920 1080
+  if [ -f "$RESOLUTION_FILE" ] ; then
+    sed "s/ //g" "$RESOLUTION_FILE" | grep "\S" | grep -vE "^\#" | while read -r line
```
dgsiegel, line 70: I am pondering about the right format of that file. One might want to use `raw` instead of `mjpeg` or even `mp4` if uvc-gadget gets support for that. In that case it would be better to add the format to that file as well. Also, it's quite error prone to parse it like that.

How about:

```
mjpeg 640 360
mjpeg 640 480
...
```
That way you could parse it with awk and drop that line if it does not contain out of 3 words. (Bonus point if you check for valid formats and if the latter two arguments are integers)

This would then for example allow to easily use `raw` etc. As for the configfs name you'd then just use the first letter of the format.

Further opinions welcome!

board/post-image.sh
westfeld: I could do that, however the aim of this PR was to make the output resolutions configurable without having to build the whole image.

One could put the default file into `/etc` and make it possible to overwrite it with another file in `/boot` however I see no advantage in there.

package/piwebcam/multi-gadget.sh
westfeld: I will have a look at this. I would then rename the input file from `resolutions.txt` to `output_formats.conf`.

board/post-image.sh
dgsiegel: Well, we get a new build with each PR/commit so this shouldn't be an issue :)

The advantage would be that configuration settings shouldn't belong inside shell files. Furthermore /boot is writeable, whereas /etc is not.

package/piwebcam/multi-gadget.sh
dgsiegel: Let's keep the "txt" extensions as all the other files do use that too. Looking forward to your ideas.

package/piwebcam/video_formats.txt
```diff
@@ -0,0 +1,16 @@
+# Default Show-me webcam supported resolutions
+#
+# format: video format(mjpeg|raw|mp4) x-dimension y-dimension
```
dgsiegel, line 3: mp4 is not yet supported unfortunately, see https://github.com/peterbay/uvc-gadget/issues/8

Currently we only have `mjpeg` and `uncompressed`, see: https://github.com/raspberrypi/linux/blob/rpi-5.4.y/drivers/usb/gadget/function/uvc_configfs.c#L767

board/post-image.sh
```diff
@@ -90,6 +90,9 @@ __EOF__
 			cat "$BR2_EXTERNAL_PICAM_PATH/camera.txt" >> "${BINARIES_DIR}/camera.txt"
 		fi
 
+		# Add default video_formats.txt
+                cp "$BR2_EXTERNAL_PICAM_PATH/package/piwebcam/video_formats.txt" "${BINARIES_DIR}/video_formats.txt"
```
dgsiegel, line 94: This should be part of `package/piwebcam/piwebcam.mk`. Any reason why you chose `video_formats` instead of `resolutions`. Happy with both, it's just a longer name.

package/piwebcam/multi-gadget.sh
```diff
@@ -64,17 +66,13 @@ EOF
...
-  config_frame mjpeg m  800  600
-  config_frame mjpeg m 1024  768
-  config_frame mjpeg m 1280  720
-  config_frame mjpeg m 1280  960
-  config_frame mjpeg m 1440 1080
-  config_frame mjpeg m 1536  864
-  config_frame mjpeg m 1600  900
-  config_frame mjpeg m 1600 1200
-  config_frame mjpeg m 1920 1080
+  grep -E "(mjpeg|raw|mp4)\s+[[:digit:]]+\s+[[:digit:]]+" $VIDEO_FORMATS_FILE | sed -E "s/\s+/ /g" | while read -r line
```
dgsiegel, line 69: Looks a lot better. Might want to add `^` to start at the beginning of a line and/or exclude lines starting with `#`. awk might be perfect for this but happy to keep it like this as well.

package/piwebcam/multi-gadget.sh
```diff
@@ -64,17 +66,13 @@ EOF
...
+  do
+    VIDEO_FORMAT=$(echo "$line" | cut -d ' ' -f1)
+    X=$(echo "$line" | cut -d ' ' -f2)
+    Y=$(echo "$line" | cut -d ' ' -f3)
+    config_frame "$VIDEO_FORMAT" m "$X" "$Y"
```
dgsiegel, line 74: "m" should be the first character of the format instead of being hardcoded

package/piwebcam/multi-gadget.sh
westfeld: resolved, see fb0418a

package/piwebcam/video_formats.txt
westfeld: I corrected that in fb0418a

board/post-image.sh
westfeld: My idea of the fix was to be able to configure the piwebcam to ignore resolutions after the image has been burned to an SD card.

So if you see that the resolution is too low, you can put the SD card into your PC and change the `video_formats.txt` file on the `/boot` partition, reinsert it back into your raspberrypi and it will work.

The reason why I changed the name was that now not only the resolution gets configurable, but also the streaming format. So I thought naming the file just `resolutions.txt` is misleading.

package/piwebcam/multi-gadget.sh
westfeld: To be honest I do not know what the "m" parameter in the config file function call does. It creates some file for USB configuration. So does it have to be changeable? How should the config file look like?

    m mjpeg 1920 1080

or

    mmjpeg 1920 1080

In the second case I would just take the very first character. What about the syntax checking? Should it be only "m" at the moment?

package/piwebcam/multi-gadget.sh
westfeld: To be honest, I have not dived into awk, yet. If it is really beneficial I would go that route, though.

package/piwebcam/multi-gadget.sh
dgsiegel: That single character is the specific streaming header descriptor, "m" for mjpeg and "u" for uncompressed. I *think* you could use any letter, hence I've said to just use the first letter of the format.

board/post-image.sh
dgsiegel: Right, this approach is sensible. What I meant however was to use the Makefile instead of the shell script to add that file.

westfeld: I have looked into this a bit more. But I did not find out how to place files in the `/boot` partition out of a buildroot package. I understood how to place files into the staging area to be included in the image, but I am not sure if this counts for the `/boot` partition, as well...

dgsiegel: Oh you don't have to, /etc would be the right place for this. A user should then be able to overwrite that file from /boot. Just add this line to `package/piwebcam/piwebcam.mk`

```
$(INSTALL) -D -m 644 $(PIWEBCAM_PKGDIR)/video_formats.txt $(TARGET_DIR)/etc/video_formats
```

Nevertheless I think we should come up with a better name for this file?

westfeld: I have implemented the installation in 618e516 and I am open to suggestions for another file name.

dgsiegel (dismissed):

dgsiegel:
@htruong any idea for a better file name? Other than that I think we can merge this now.

westfeld:
What do you think of ’vformats.txt’

htruong:
I think `video_formats.txt` is good (`vformats.txt`, on the other hand is short but is a bit obscure). I will give it a spin before merging :)

westfeld:
I thought adding some documentation would be nice. Is it OK to put it there or would it be better in the wiki?

dgsiegel:
> I thought adding some documentation would be nice. Is it OK to put it there or would it be better in the wiki?

This is great, thanks! I'd only add that client apps will chose resolutions automatically, hence it's only needed to reduce available resolutions if you want to enforce a specific setting.

westfeld:
@dgsiegel Thanks for the kind introduction and the comments. It helped me as a first-contributor to this project. I am using the showmewebcam everyday and it is great (Pi0 + HQ).

dgsiegel:
> @dgsiegel Thanks for the kind introduction and the comments. It helped me as a first-contributor to this project. I am using the showmewebcam everyday and it is great (Pi0 + HQ).

Thanks :) Glad to have you on board. Let's wait for @htruong to give it a spin and then merge it.

README.md
```diff
@@ -153,6 +153,55 @@ You can edit `camera.txt` on-target by remounting `/boot` read-write:
...
 
+### Configure available camera resolutions and streaming format
+
+The camera advertizes to the connected host PC the available resolutions and
+streaming formats of the video signal. Usually the best suitable resolution is
+chosen. The available resolutions can be reduced to enforce a specific setting.
+
+For example: MS Teams on Windows 10 is picking only a very low resolution and
+the picture looks blurry. To fix this, the low resolutions can be removed from
+the advertized set.
```
dgsiegel, line 164: Just a minor nitpick: Many haven't had this issue with Teams on WIndows, some did however. I'd guess this paragraph would be better suited for the FAQ (https://github.com/showmewebcam/showmewebcam/wiki/FAQ) than the Readme.

westfeld: I have moved the use case from the README to the wiki.

package/piwebcam/multi-gadget.sh
```diff
@@ -64,17 +69,20 @@ EOF
...
+  do
+    VIDEO_FORMAT=$(echo "$line" | cut -d ' ' -f1)
+    HDR_DESC=$(echo "$VIDEO_FORMAT" | cut -c 1)
+    X=$(echo "$line" | cut -d ' ' -f2)
+    Y=$(echo "$line" | cut -d ' ' -f3)
+    config_frame "$VIDEO_FORMAT" "$HDR_DESC" "$X" "$Y"
+  done
 
   mkdir -p functions/uvc.usb0/streaming/header/h
   ln -s functions/uvc.usb0/streaming/mjpeg/m  functions/uvc.usb0/streaming/header/h
```
wherget, line 88: AFAICS, you'd need to link `functions/uvc.usb0/streaming/uncompressed/u` as well for uncompressed formats to work. Did you try it with `uncompressed`?

package/piwebcam/multi-gadget.sh
```diff
@@ -64,17 +69,20 @@ EOF
...
-  config_frame mjpeg m  800  600
-  config_frame mjpeg m 1024  768
-  config_frame mjpeg m 1280  720
-  config_frame mjpeg m 1280  960
-  config_frame mjpeg m 1440 1080
-  config_frame mjpeg m 1536  864
-  config_frame mjpeg m 1600  900
-  config_frame mjpeg m 1600 1200
-  config_frame mjpeg m 1920 1080
+  if [ ! -e $VIDEO_FORMATS_USER_FILE ] ; then
```
wherget, line 72: This sounds the wrong way round. If user file does NOT exist, use user file as formats file? Did you check this with a custom file that does specify some other formats?

Also, how do you feel about a stronger check, i.e. "file is readable" `-r` or even "not empty" `-s`

package/piwebcam/piwebcam.mk
```diff
@@ -21,6 +21,7 @@ define PIWEBCAM_INSTALL_TARGET_CMDS
 	$(INSTALL) -D -m 0755 $(@D)/uvc-gadget $(TARGET_DIR)$(PIWEBCAM_DEST_DIR)
 	$(INSTALL) -D -m 0755 $(PIWEBCAM_PKGDIR)/multi-gadget.sh $(TARGET_DIR)$(PIWEBCAM_DEST_DIR)
 	$(INSTALL) -D -m 0755 $(PIWEBCAM_PKGDIR)/start-webcam.sh $(TARGET_DIR)$(PIWEBCAM_DEST_DIR)
+	$(INSTALL) -D -m 644 $(PIWEBCAM_PKGDIR)/video_formats.txt $(TARGET_DIR)/etc/video_formats.txt
```
wherget, line 24: If only for symmetry reasons, I'd prefer the full set of mode bits, i.e. `0644`.

wherget (commented):
I really like this (tbh I implemented almost the same, but yours has a few more niceties), but I think there's still a few kinks in the code, which I've taken the liberty to annotate.

package/piwebcam/multi-gadget.sh
dgsiegel: Oh true, I've missed that. Thanks!

@westfeld, basically for every given format you need to link the headers

For mjpeg:

`   ln -s functions/uvc.usb0/streaming/mjpeg/m  functions/uvc.usb0/streaming/header/h `

For uncompressed:

`   ln -s functions/uvc.usb0/streaming/uncompressed/u  functions/uvc.usb0/streaming/header/h `

package/piwebcam/multi-gadget.sh
dgsiegel: Right, you could set ` VIDEO_FORMATS_FILE=/etc/video_formats.txt` as default, and overwrite it if a user file exists. It should always be readable, as /boot is FAT. But you're right, better safe than sorry :)

htruong (commented):
Also, can you kindly squash those commits together and then force push to the PR :)? That will make it a little bit cleaner.

westfeld:
I have implemented the changes locally, but I have to test the resolutions and the handling of the uncompressed video before pushing, first.

HunterAP23:
Saw that the issue I created was tagged in this, thanks for working so hard on this!

I wanted to ask if higher than 1080p resolutions were tested, as well as >30fps was also tested (or if they need to be tested). I know that #72 is a not-yet merged request or adding 60fps support, but I was curious to see if it was at least technically feasible. 

The Pi Zero / Zero W could theoretically support 1080p60 over USB 2 (30fps is 1.87 Gbps and 60fps would be 3.73 Gbps, which is below the 5Gbps spec of USB 2).

The Pi 4 has the advantage of having 10Gbps bandwidth, so it could achieve 3840x2160 at 30fps using 7.46 Gbps, or an odd resolution like 1440p at 60fps using 6.64 Gbps. 2160p would require the HQ camera module, whereas 1440p would requirethe Camera V2 module.

I can try and test these settings myself and reporting back if need be, I was just curious if there are plans to add those options or if there are any known issues for them

westfeld:
Thanks @HunterAP23 for commenting. I will get back to you once I get this PR up and running.

Right now I have a HQ camera and a Pi zero and a Pi zero w for testing.

I will test several resolutions and streaming formats (mjpeg vs. uncompressed).

Changing the frame rate is at the moment not yet implemented, but it is certainly possible. I think it is important to only allow settings which are compatible to the used hardware.

htruong (approved):

package/piwebcam/multi-gadget.sh
```diff
@@ -64,27 +69,29 @@ EOF
...
-}
 
+   grep -E "^(mjpeg|uncompressed)\s+[[:digit:]]+\s+[[:digit:]]+" $FORMATS_FILE | sed -E "s/\s+/ /g" | while read -r line
+  do
+    VIDEO_FORMAT=$(echo "$line" | cut -d ' ' -f1)
+    HDR_DESC=$(echo "$VIDEO_FORMAT" | cut -c 1)
+    X=$(echo "$line" | cut -d ' ' -f2)
+    Y=$(echo "$line" | cut -d ' ' -f3)
+    config_frame "$VIDEO_FORMAT" "$HDR_DESC" "$X" "$Y"
+    ln -sf functions/uvc.usb0/streaming/$VIDEO_FORMAT/$HDR_DESC  functions/uvc.usb0/streaming/header/h
```
wherget, line 92: Shellcheck reasonably complained about quoting here
```suggestion
    ln -sf "functions/uvc.usb0/streaming/$VIDEO_FORMAT/$HDR_DESC"  functions/uvc.usb0/streaming/header/h
```

htruong (commented):
Squash with previous commit?

package/piwebcam/multi-gadget.sh
```diff
@@ -64,27 +69,29 @@ EOF
...
-  config_frame mjpeg m 1280  960
-  config_frame mjpeg m 1440 1080
-  config_frame mjpeg m 1536  864
-  config_frame mjpeg m 1600  900
-  config_frame mjpeg m 1600 1200
-  config_frame mjpeg m 1920 1080
+  if [ -r $VIDEO_FORMATS_USER_FILE ] ; then
+      FORMATS_FILE=$VIDEO_FORMATS_USER_FILE
+  else
+      FORMATS_FILE=$VIDEO_FORMATS_FILE
```
dgsiegel, line 75: Small nitpick: the file uses 2 spaces indentation, here's 4.

package/piwebcam/multi-gadget.sh
```diff
@@ -64,27 +69,29 @@ EOF
...
   mkdir -p functions/uvc.usb0/streaming/header/h
-  ln -s functions/uvc.usb0/streaming/mjpeg/m  functions/uvc.usb0/streaming/header/h
   ln -s functions/uvc.usb0/streaming/header/h functions/uvc.usb0/streaming/class/fs
   ln -s functions/uvc.usb0/streaming/header/h functions/uvc.usb0/streaming/class/hs
   ln -s functions/uvc.usb0/control/header/h   functions/uvc.usb0/control/class/fs
 
   ln -s functions/uvc.usb0 configs/c.2/uvc.usb0
-}
 
+   grep -E "^(mjpeg|uncompressed)\s+[[:digit:]]+\s+[[:digit:]]+" $FORMATS_FILE | sed -E "s/\s+/ /g" | while read -r line
```
dgsiegel, line 85: Small nitpick: single space before `grep`

package/piwebcam/multi-gadget.sh
```diff
@@ -64,27 +69,29 @@ EOF
...
+  do
+    VIDEO_FORMAT=$(echo "$line" | cut -d ' ' -f1)
+    HDR_DESC=$(echo "$VIDEO_FORMAT" | cut -c 1)
+    X=$(echo "$line" | cut -d ' ' -f2)
+    Y=$(echo "$line" | cut -d ' ' -f3)
+    config_frame "$VIDEO_FORMAT" "$HDR_DESC" "$X" "$Y"
+    ln -sf "functions/uvc.usb0/streaming/$VIDEO_FORMAT/$HDR_DESC"  functions/uvc.usb0/streaming/header/h
```
dgsiegel, line 92: Small nitpick: double space between the two arguments

package/piwebcam/video_formats.txt
```diff
@@ -0,0 +1,16 @@
...
+mjpeg 800 600
+mjpeg 1024 768
+mjpeg 1280 720
+mjpeg 1280 960
+mjpeg 1440 1080
+mjpeg 1536 864
+mjpeg 1600 900
+mjpeg 1600 1200
+mjpeg 1920 1080
+
```
dgsiegel, line 16: Small nitpick: Empty line

dgsiegel (changes requested):
Just a few small nitpicks here and there. Other than that I am looking forward to our next release :)

whi-tw:
Is there any way to speed this along if it appears to just be blocked by style changes (ie. fix the formatting during the merge process?)

HunterAP23:
I wasn't sure how to go about making the changes requested, so I opted to make a fork of [westfeld:select_resolutions](https://github.com/westfeld/showmewebcam/tree/select_resolutions) with those changes made and created a [PR ](https://github.com/westfeld/showmewebcam/pull/1) to it. Once that PR is merged, those changes should appear on this PR.

dgsiegel:
I've updated and rebased this PR in #154

dgsiegel:
Merged via #154. Thanks @westfeld, @HunterAP23 & @wherget for your help!

PR #134, opened 2021-04-28
update uvc-gadget to e9a733f

dgsiegel:


htruong (approved):

htruong:
nit. Would be nice to have s short summary of what is changed and why the uprev is beneficial.

PR #147, opened 2021-06-28
[WIP] Add support for rpi3

towolf:
I tried to add support for Raspberry PI 3B, but I only got a message `brcmf_sdio_htclk: HT Avail timeout`. Then I proceeded to disable WLAN in DeviceTree overlays, but then the Pi only stopped printing that message and still did not boot up. I only got a blinking cursor.

Not sure how else the build options should be modified to make it boot. So I would need some advice at this point.

PXL_20210624_135645281


htruong:
Hmm, how do you get the rpi3 to appear as a gadget?

htruong:
the blinking cursor is likely not the problem. The services started just fine, it hangs there because there is no login prompt service that starts on the default screen console. You can use the raspberry pi serial headers to see if you get a login prompt.

Leapo:
> 
> 
> Hmm, how do you get the rpi3 to appear as a gadget?

The Pi 3A and 3A+ support gadget mode, but I'm not sure how you'd make this work with a 3B or 3B+

towolf:
Hm, maybe I was mistaken about how this can work. I never got that far. 

Are you saying that I cannot physically connect as a USB gadget?

Leapo:
> Are you saying that I cannot physically connect as a USB gadget?

Gadget Mode is only supported when the SoC's onboard USB port is directly exposed.

The Pi 3B and Pi 3B+ have the SoC's onboard USB port connected to a combination USB Hub and Ethernet Controller chip, which then provides the 4 USB 2.0 ports and ethernet port along the back edge of the Pi. Gadget mode cannot pass through this additional chip.

The Pi 3A and Pi 3A+, on the other hand, DO NOT have a USB hub or ethernet adapter connected to the SoC. They provide a single USB 2.0 port at the edge of the board, with support for gadget mode.

If you want to support the Pi 3, it'll have to be the 3A and 3A+

PR #154, opened 2021-08-18
Configurable resolutions

dgsiegel:
Includes the following changes:

- use [:space:] instead of \s as older versions of grep don't support \s
- remove sed as the busybox version uses different flags than gnu sed
- use awk to parse the resolutions which is less error-prone
- call config_frame before setting up streaming headers as the setup will break otherwise
- update buildroot to 2021.02.4

This replaces PR #128


htruong (approved):
Thanks for doing this. Looks clean to me.

PR #156, opened 2021-09-01
set default uvc host controls to brightness, contrast and saturation

dgsiegel:


htruong:
Sorry for getting to this too late, this PR https://github.com/showmewebcam/showmewebcam/pull/167 also fixes the doc as well, so I think we should close this to merge the other one :)

PR #159, opened 2021-09-14
Add sha256 checksum file to releases

### LINKED ISSUE #157: Add file checksums to releases
Hello all.

I've updated the release workflow to add a CHECKSUMS file. The file would be available on the releases page, with the build artifacts.

I'm raising the issue here in order to gather feedback if this is an welcome change/improvement. I'll raise the PR if I have a positive response.

channelbeta:
Closes #157 

This PR adds a step that creates a CHECKSUMS file to be uploaded with other assets on release. I do it at the end of the workflow because then we have all assets already present in the same working directory.

My first iteration was using sha512, but I decided to replace it with sha256 for no particular reason.

Please feel free to make any suggestions.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -128,9 +128,13 @@ jobs:
         with:
           name: release
 
+      - name: Create checksums file
+        run: |
+          sha256sum "sdcard*" > CHECKSUMS.sha256
+
       - name: Upload Binaries to Release
         uses: alexellis/upload-assets@0.2.2
         env:
           GITHUB_TOKEN: ${{ github.token }}
         with:
-          asset_paths: '["sdcard*"]'
+          asset_paths: '["sdcard*", "CHECKSUMS.sha256"]'


dgsiegel:
This looks great, thanks!

Just one question: Did you do any research on what the "correct" filename would be? I've seen some projects using only `CHANGELOG` or `CHANGELOG.txt`.

dgsiegel (approved):

channelbeta:
Hi @dgsiegel , I haven't done any research regarding the naming scheme.

According to Wikipedia (but without citing any sources to the claim):
https://en.wikipedia.org/wiki/File_verification

> The particular hash algorithm used is often indicated by the file extension of the checksum file. 

While I prefer this style of `NAME.algorithm`, I can change it if you'd like :smiley:

dgsiegel:
Couldn't find any standard either, so let's leave it like this. Thanks :)

PR #165, opened 2021-11-01
readme: Add links to known issues with macOS

htruong:


htruong:
No longer needed, we found the solution.

PR #167, opened 2021-11-03
Enable only controls that the RPi camera supports

friissoren:
The values that are currently set by default in the patch for updating the f_uvc.c kernel driver file will try to enable some controls that are not supported by the 8 Mpixel (v2) camera on the RaspberryPi 4B HW and leaves out some that are supported.

Fix the values to match exactly the controls that are possible.

I don't have any other HW than the RPi 4B + v2 camera so I don't know if these settings are completely valid on other HW combinations.

NOTE: In order to get the exposure and white balance controls working, fixes are needed also in the uvc-gadget code.
https://github.com/peterbay/uvc-gadget/pull/23

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,9 +167,15 @@ You can add the parameters to `cmdline.txt` on the boot volume as follows:
 usb_f_uvc.camera_terminal_controls=0,0,0 usb_f_uvc.processing_unit_controls=0,0
 ```
 
-The previous example sets all controls to disabled, and should thus be safe.
+The above example disables all controls and should thus be safe.
 The parameters directly correspond to the `bmControls` bitfields in the descriptor.
 Please, again, read the documentation linked above.
+
+An example for enabling all the available controls for the 8 Mpixel (v2) camera for the Raspberry Pi 4B is given below:
+
+```
+usb_f_uvc.camera_terminal_controls=10,0,0 usb_f_uvc.processing_unit_controls=219,4
+```
 
 ### Configure available camera resolutions and streaming format
 
--- a/patches/linux-custom/0001-linux-enable-more-video-controls.patch
+++ b/patches/linux-custom/0001-linux-enable-more-video-controls.patch
@@ -6,19 +6,19 @@ index fb0a892..5c850be 100644
  module_param_named(trace, uvc_gadget_trace_param, uint, 0644);
  MODULE_PARM_DESC(trace, "Trace level bitmask");
  
-+unsigned int uvc_gadget_camera_terminal_controls[3] = {2, 0, 0};
++unsigned int uvc_gadget_camera_terminal_controls[3] = {10, 0, 0};
 +module_param_array_named(camera_terminal_controls, \
 +		uvc_gadget_camera_terminal_controls, \
 +		uint, NULL, 0644);
 +MODULE_PARM_DESC(camera_terminal_controls, \
-+		"Camera terminal control availabilty bitfield");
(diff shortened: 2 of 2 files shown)


htruong:
Thanks for doing this! Do you mind having some explanation to the magic values?

friissoren:
Those values are explained [here](https://github.com/showmewebcam/showmewebcam#overriding-the-controls-available-to-the-host-computer) and [here](https://github.com/showmewebcam/showmewebcam/wiki/UVC-Controls).

The following assumes you are using a Linux (e.g. Ubuntu) machine as the host for debugging.

If you log into the RPi web camera ` sudo picocom -b 115200 /dev/ttyACM0` and do a list of the V4L2 controls that are supported by the RaspberryPi ISP/camera driver: `v4l2-ctl -d 0 --all`, and then compare that against the "magic" bitmap values that are currently configured in the patch, it is clear there is a mismatch.

This can also be seen clearly from the various errors when following the dmesg log after enabling additional traces for the uvc driver:
```
sudo su
echo 0xffff > /sys/module/uvcvideo/parameters/trace
exit
dmesg -w
Now plug-in the USB cable connecting the RaspberryPi to the PC.
```
Remember to `echo 0x0` back into that trace parameter when you are done (or reboot). Otherwise it keeps logging and I at some point found myself having a 35 Gbyte log file in /var/log.

If you look carefully at the list of controls that v4l2-ctl gives, you can see that there are additional controls available that do not have any clear match in the UVC interface. In principle also those could be mapped to some of the controls defined in the UVC specification but they wouldn't really match the names and it would be quickly confusing to use those on the host PC since the names in whatever UI application you use would be misleading. I did this already for the white-balance control in the uvc-gadget [PR](https://github.com/peterbay/uvc-gadget/pull/23) since I personally had a need to control the white-balance and this was the only way to do it. It is a bit of a hack though.

htruong (approved):
Looks good to me!

friissoren:
Thanks for merging.

Just remember that the exposure and white balance control will not work unless you have the uvc-gadget updated with the changes from PR https://github.com/peterbay/uvc-gadget/pull/23 (or https://github.com/peterbay/uvc-gadget/pull/24 if you use the new code line).

dgsiegel:
@friissoren I am a bit confused about those specific values. In this discussion we figured a better setting would be `0,0,0` and `11,0`. However I don't have a v2 camera to test. Do these values work for you as well?

https://github.com/showmewebcam/showmewebcam/issues/116#issuecomment-901994507

friissoren:
With the values that were committed from this PR, the following controls are available on Linux and Windows (assuming you include also the uvc_gadget [PR#23](https://github.com/peterbay/uvc-gadget/pull/23)):
- Brightness
- Contrast
- Saturation
- White Balance Blue Component (this doesn't appear to do anything though. Neither when doing it directly from V4L2)
- Power Line Frequency
- White Balance Temperature (this one is a bit of a hack since it now selects between preset temperatures but doesn't describe those in the UI)
- Sharpness
- Exposure Auto/Manual
- Exposure Absolute

I tested the `0,0,0` + `11,0` settings on the RPi4 + 8Mpixel v2 camera HW. The following controls were available then:
- Brightness
- Contrast
- Saturation

It is possible that some of the controls that are now enabled with the `10,0,0` + `219,4` values are not available for the 5 or 12 Mpixel cameras or the RPi Zero. I unfortunately don't have any of that HW and cannot test.

Issue #164, opened 2021-10-27
No webcam is detected after updating to macOS 12.0.1


bl1nk:
Today I updated two of my MacBooks to macOS 12.0.1. When connecting my showmewebcam-equipped Raspberry Pi Zero W no devices are detected in any application. This happened on an Intel MacBook Pro and a M1 MacBook Air so at least it is not related to the architecture. 

In Console.app I could find some errors that might be related.

```
default	21:00:03.358719+0200	kernel	set CameraAssistantBundleID : com.apple.cmio.uvcassistantextension
default	21:00:03.358917+0200	kernel	set CameraAssistantBundleID : com.apple.cmio.uvcassistantextension
default	21:00:03.359833+0200	kernelmanagerd	Received kext load notification: com.apple.DriverKit-UVCUserService
default	21:00:03.359627+0200	UVCAssistant	UVCExtensionProvider: [0x156f04100] Found : 0x10000c246 bundleID com.apple.cmio.uvcassistantextension
default	21:00:03.359859+0200	kernelmanagerd	Received kext load notification: com.apple.DriverKit-UVCUserService
default	21:00:03.359731+0200	UVCAssistant	UVCExtensionProvider: [0x156f04100] VID 1d6b PID : 104 AppleCameraSupport No
default	21:00:03.359816+0200	UVCAssistant	UVCExtensionProvider: [0x156f04100] Association ID for service 10000c246 : 0
default	21:00:03.574901+0200	UVCAssistant	UVCExtensionDevice:0x0 [0x156e18de0] Association ID for service 10000c246 : 0
default	21:00:03.575129+0200	UVCAssistant	UVCExtensionDevice:0x0 [0x156e18de0] LocalizedDeviceName Piwebcam manufacturerName Show-me Webcam Project UUID 00000000-0224-2000-1D6B-000001040000
default	21:00:03.575254+0200	UVCAssistant	UVCExtensionDevice:0x0 [0x156e18de0] Post PowerLog {
    "VDCAssistant_Device_GUID" = "00000000-0224-2000-1D6B-000001040000";
    "VDCAssistant_Power_State" = Off;
}
default	21:00:03.575395+0200	UVCAssistant	boot args [enableCameraExtension : 0]
default	21:00:03.575457+0200	UVCAssistant	UVCUSBDevice: 0x0 [0x16010cef0] -[UVCUSBDevice initWithInterface:queue:streamDataHandler:stateNotificationHandler:]
[...]

PR #169, opened 2021-11-07
Use c.1 config container instead of c.2. Fixes issue #164

stoggi:
As per @tomjtobin comment in showmewebcam/showmewebcam#164, `multi-gadget.sh` reports only one configuration but uses container `c.2`.
This causes macOS 12.0.1 to no longer recognise the UVC Camera.

Changing to `c.1` works on macOS 12.0.1 and 11.5.2.
Not tested on other systems.

More information on UVC gadget configuiration: https://www.kernel.org/doc/Documentation/usb/gadget_configfs.txt


htruong:
Thanks for doing this! I have tested this on my laptop and it worked!

htruong (approved):

PR #188, opened 2022-02-04
improve build script

dgsiegel:
This PR does the following:

- Gets the buildroot commit id automatically instead of having to change release.yml every time we change it
- Properly set os-release file with the current showmewebcam version id
- Fixes the sha265sum command
- Uses the full git version description for image file names (ie. sdcard-raspberrypi0-v1.91-12-gffc2ff3.img)
- Switch to the official buildroot repository

### DIFF
--- a/board/post-build.sh
+++ b/board/post-build.sh
@@ -33,3 +33,14 @@ ln -sf /dev/null "${TARGET_DIR}"/etc/systemd/system/systemd-update-utmp-runlevel
 if [[ ${TARGET_DIR} != *"raspberrypi0cam"* ]]; then
   ln -sf /dev/null "${TARGET_DIR}"/etc/systemd/system/network.service
 fi
+
+# Set os-release file
+SHOWMEWEBCAM_VERSION=$(support/scripts/setlocalversion "${BR2_EXTERNAL_PICAM_PATH}")
+
+cat > "${TARGET_DIR}"/usr/lib/os-release <<EOF
+NAME="Show-me webcam"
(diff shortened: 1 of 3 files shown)


dgsiegel:
Don't merge yet, still need to fix `/etc/os-release`

board/post-build.sh
```diff
@@ -33,3 +33,15 @@ ln -sf /dev/null "${TARGET_DIR}"/etc/systemd/system/systemd-update-utmp-runlevel
...
+
+# Set os-release file
+SHOWMEWEBCAM_VERSION=$(support/scripts/setlocalversion "${BR2_EXTERNAL_PICAM_PATH}")
+
+cat > "${TARGET_DIR}"/usr/lib/os-release <<EOF
+NAME="Show-me webcam"
+VERSION=${SHOWMEWEBCAM_VERSION}
+ID=showmewebcam
+VERSION_ID=${SHOWMEWEBCAM_VERSION}
+BUILD_ID=$(date +%Y.%m.%d)-${SHOWMEWEBCAM_VERSION}
```
sftim, line 45: I'm a fan of [reproducible builds](https://wiki.debian.org/ReproducibleBuilds).
How about using the commit date and time for `HEAD` instead?

dgsiegel: You are absolutely right. However it's quite hard to get the actual commit date at that time. So I've just removed that line as it is optional anyway and we've got the same information in `VERSION_ID` already.

htruong (approved):
Thanks! Everything looks good to me.

Issue #183, opened 2022-01-16
Restart uvc-gadget on exit


wherget:
(This is a report submitted by [vogarth] on Discord)

## Expected Behavior

When a resolution request or issue with the upstream (towards sensor) v4l / mmal layer causes `uvc-gadget` to exit, it is restarted, making it possible to restart a camera session or retrieve log output.

## Current Behavior

Once uvc-gadget has exited, showmewebcam remains in an unusable state. Power-cycling the pi is required to be able to restart capture.

## Possible Solution

(suggestion by [vogarth])
A simple reset camera / restart app loop around uvc-gadget in the startup script would achieve this.

(my comment)
Might just be a case of setting the appropriate `Restart` flag on `uvc-webcam.service`. Currently seems to use the default of `Restart=no`.

I would like to posit for discussion: Would it be surprising to reset the camera parameters in such an event (as it is currently the case)? I feel like I'd expect the sensor parameters to remain as is, if that is possible. That may mean separating the "apply `camera.txt`" logic into another separate service that's only run on boot. OTOH, if ending the capture session abnormally leaves the sensor in an undefined state, it might still be advisable to reset on restart.

dgsiegel:
I guess using the right restart/onfailure flags in the systemd service would be enough. Wanna provide a PR?

> I would like to posit for discussion: Would it be surprising to reset the camera parameters in such an event (as it is currently the case)? I feel like I'd expect the sensor parameters to remain as is, if that is possible.

Fully agree.

PR #190, opened 2022-02-06
Tweak systemd unit definition

sftim:
:information_source: UNTESTED

These changes ought to make the systemd unit automatically restart (see #183), and run with extra security restrictions.
I haven't tried them though. Nonetheless I hope the PR is useful. Anyone who wants to is welcome to test this out and report back.


dgsiegel:
Looks good! We might even make use of more systemd capabilites. Looking at `start-webcam.sh`, it could be easily replaces as it does three things:

1. Check if both `/dev/video0` and `/dev/video1` exist. Could be replaced with `ConditionPathExists`
2. Applying config. Could be a small shell script run with `ExecStartPre`
3. Running the actual uvc-gadet. Could be run with `ExecStart`

dgsiegel:
@sftim wanna check these three things out?

sftim:
I'm not going to have any time to move this PR forward with extra changes, at least not for the foreseeable. Sorry about that.
I'm happy if someone wants to open a new PR that takes my 3 commits and adds some more, or to start from scratch.

sftim:
Whoops.

sftim:
Even if this isn't perfect, perhaps it's OK to merge as-is? The key thing would be to verify it actually works.

dgsiegel:
> Even if this isn't perfect, perhaps it's OK to merge as-is? The key thing would be to verify it actually works.

I'd rather not merge something that wasn't verified. But you can easily use the Github Action to build your images and try those.

PR #197, opened 2022-05-03
Buildroot update

dgsiegel:


kbingham:
Not fully sure what's going on with the version issue on the libcamera patch, but if that works that's enough for now. The version probably only affects the 'string' that reports the version compiled. A nice to have but not essential.

What effect does all of this update have on the /boot/config.txt ? Does it automatically enable the libcamera stack? or is the legacy enabled by default in this buildroot?

kbingham:
I also see updates to the kernel configs. Are there any changes there that could affect how ConfigFS is configured or enabled?

kbingham:
So contrary to what I thought last time - there is access to the hardware jpeg encoder, through the V4L2 M2M device node. (/dev/video31 by default by the looks of it - but we shouldn't rely on that number).

Support for this should be added to either libcamera or uvc-gadget ... and then I hope/expect we'll get back to the existing performance capabilities!.

If anyone has the ability to help work on this let me know - otherwise I'll see if we can look at it when we have some downtime.

ventilator:
Will this also support the new camera module 3?

bbdolly:
@kbingham @dgsiegel 
I hope you are doing well. I have both the Raspberry Pi Zero and the Zero 2 W. A few days ago, I successfully run this project  using the Raspberry Pi Zero and now I want to replicate it on my Raspberry Pi Zero 2 W.

According to previous comments, this PR will support the Zero 2 W. However, I'm unsure about how to build it for the Zero 2 W. The available targets I can select are raspberrypi0, raspberrypi0w, and raspberrypi4. Could you please advise me on which one I should use for the Zero 2 W?

Thanks,
Xue

PR #209, opened 2022-09-19
package: use HTTPS link instead of GIT to download packages

GDGooD:
Unauthorized git:// links are no more supported by Github. Use https:// instead.
https://www.infoq.com/news/2021/09/github-improves-security/


dgsiegel (approved):
Looks good to me!

PR #227, opened 2024-01-04
Update README.md

NKT00:
Added a getting started bit for Windows and the Pi Zero, plus some troubleshooting advise.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,8 +35,27 @@ Show-me webcam is proudly powered by [peterbay's uvc-gadget](https://github.com/
 
 ## Instructions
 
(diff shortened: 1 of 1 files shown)


dgsiegel:
Happy to update the link, but I guess everything else should me moved to the wiki. Wanna do that?

NKT00:
I've now found the wiki. I'll do an update there tomorrow or something. The Windows stuff should be in the Readme alongside the Mac and Linux instructions though, surely?

dgsiegel:
> I've now found the wiki. I'll do an update there tomorrow or something. The Windows stuff should be in the Readme alongside the Mac and Linux instructions though, surely?

Yeah, that would be great!

